PR #74, opened 2021-03-06
should patch webhook on managed cluster

ycao56:
https://github.com/open-cluster-management/backlog/issues/10074
https://github.com/open-cluster-management/backlog/issues/10096
https://github.com/open-cluster-management/backlog/issues/10131
https://github.com/open-cluster-management/backlog/issues/10139
https://github.com/open-cluster-management/backlog/issues/10142


gparvin:
I think https://github.com/open-cluster-management/backlog/issues/10181 hit the same issue prior to the fix getting in.

PR #77, opened 2021-03-09
since ignore is set namespace creation works

gparvin:
Canary failure in issue https://github.com/open-cluster-management/backlog/issues/10253 I think happened
because now we set an ignore flag to ignore gatekeeper violations.  So in this case the namespace still gets
created, but the audit entry is still logged.

May want @ycao56 to decide if this is the way we want to keep this test.


gparvin:
This problem didn't happen in the latest canary.  The tests ran in a different order so I'm not convinced this fix is appropriate yet.

gparvin:
Going to close this for now.  I don't want to introduce additional breaks since the last run was ok.

PR #82, opened 2021-03-17
increase timeout for gatekeeper pod to be running

gparvin:
I think this would be the proposed change to address the failing canary.  https://github.com/open-cluster-management/backlog/issues/10545

@ycao56 What do you think about the canary failure and whether you think this will help?


ycao56 (approved):

ycao56:
yes, I agree. The rest of the case all passed meaning it was just not enough time for pod to be ready

PR #115, opened 2021-06-29
Skip compliance operator tests on ROSA

JustinKuli:
A configuration policy cannot create the `openshift-compliance` namespace on ROSA clusters because `openshift-*` namespaces are restricted to "non-customer" workloads. Since that is the correct namespace for the operator, we will just skip those tests on clusters with this situation.

This also includes a fix for https://github.com/open-cluster-management/governance-policy-framework/pull/114, because I forgot to update the gatekeeper namespace again after changing it in the policy. It is now `gatekeeper-system`.

Update:
After getting my hands on a ROSA cluster to test the policies more throughly, I found that the upstream operator can install correctly with these namespace changes, but it won't function correctly because it would still need to create the `openshift-gatekeeper-system` namespace to run the actual gatekeeper pods. So, this now skips the upstream gatekeeper test on ROSA clusters (and others where those namespaces might be restricted).

The downstream gatekeeper operator does still work on ROSA (somehow).

### DIFF
--- a/test/policy-collection/policy_gatekeeper_operator_test.go
+++ b/test/policy-collection/policy_gatekeeper_operator_test.go
@@ -38,13 +38,13 @@ var _ = Describe("", func() {
 			Skip("Skipping as upstream gatekeeper operator requires the ability to create the openshift-gatekeeper-system namespace")
 		}
 	})
-	const gatekeeperPolicyURL = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/master/community/CM-Configuration-Management/policy-gatekeeper-operator.yaml"
+	const gatekeeperPolicyURL = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/main/community/CM-Configuration-Management/policy-gatekeeper-operator.yaml"
 	const gatekeeperPolicyName = "policy-gatekeeper-operator"
-	const GKPolicyYaml = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/master/community/CM-Configuration-Management/policy-gatekeeper-sample.yaml"
+	const GKPolicyYaml = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/main/community/CM-Configuration-Management/policy-gatekeeper-sample.yaml"
 	const GKPolicyName = "policy-gatekeeper"
-	const GKAssignPolicyYaml = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/master/community/CM-Configuration-Management/policy-gatekeeper-image-pull-policy.yaml"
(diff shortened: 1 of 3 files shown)


dhaiducek (approved):

test/policy-collection/policy_gatekeeper_operator_test.go
```diff
@@ -133,13 +133,13 @@ var _ = Describe("", func() {
 		It("Gatekeeper operator pod should be running", func() {
 			By("Checking if pod gatekeeper-operator has been created")
 			Eventually(func() interface{} {
-				podList, err := clientManaged.CoreV1().Pods("gatekeeper-operator").List(context.TODO(), metav1.ListOptions{LabelSelector: "control-plane=controller-manager"})
+				podList, err := clientManaged.CoreV1().Pods("gatekeeper-system").List(context.TODO(), metav1.ListOptions{LabelSelector: "control-plane=controller-manager"})
 				Expect(err).To(BeNil())
 				return len(podList.Items)
 			}, defaultTimeoutSeconds*8, 1).ShouldNot(Equal(0))
 			By("Checking if pod gatekeeper-operator is running")
```
dhaiducek, line 140: Looks like the test description should change also

test/policy-collection/policy_gatekeeper_operator_test.go
```diff
@@ -544,11 +544,11 @@ var _ = Describe("", func() {
 				out, _ := utils.KubectlWithOutput("get", "ns", "openshift-gatekeeper-system", "--kubeconfig="+kubeconfigManaged)
 				return out
 			}, defaultTimeoutSeconds*4, 1).Should(ContainSubstring("namespaces \"openshift-gatekeeper-system\" not found"))
```
dhaiducek, line 546: Does this need to all be `gatekeeper-system` also or is this something else?

dhaiducek (changes requested):

test/policy-collection/policy_gatekeeper_operator_test.go
JustinKuli: This is the namespace where the gatekeeper-operator installed the "actual" gatekeeper pods. But, since noticing this and running some more tests on the ROSA cluster, it looks like the upstream gatekeeper-operator will not work on ROSA clusters. The namespace change makes it able to be installed, but not actually function (since it will still try to create this `openshift-gatekeeper-system` namespace and be denied).

I'm not sure exactly why, but the downstream gatekeeper operator *does* have permissions to make the `openshift-gatekeeper-system` namespace, so it's unaffected by this and can still run correctly on ROSA.

test/policy-collection/policy_gatekeeper_operator_test.go
JustinKuli: Updated the description for the pod name it is looking for.

dhaiducek (approved):

PR #116, opened 2021-06-29
Add case to not skip gatekeeper tests

JustinKuli:
The sample gatekeeper policy created during the tests makes an admission
webhook that prevents the server-side dry-run from testing whether an
openshift-* namespace can be created. But if gatekeeper (and this rule)
are already present, then we can assume we would be able to make the ns.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/13759


JustinKuli:
/hold

I'll wait for the e2e tests to run this time before merging

dhaiducek (approved):

JustinKuli:
/unhold

nevermind, the PR tests don't look like they run these anyway

PR #118, opened 2021-06-30
Pin version of policy-collection for 2.2

JustinKuli:
Changes to policies in the collection repo could break these tests. This
pins the version of those policies to the last commit in that repo
before the 2.2 release.


gparvin (dismissed):

JustinKuli:
It took me a bit to try to get them running locally, but I think the e2e tests are broken on the 2.2 branch for other reasons... the new commit will run a travis build to double check that I don't (still) have something messed up in my environment.

JustinKuli:
After talking with @gparvin we will address the canary failures a different way.

PR #137, opened 2021-08-17
Move common test setup to separate package

JustinKuli:
One interesting detail here is that the command line flags are parsed
by ginkgo at some unknown time. So the `clientHub` can't be defined in
the `common` package, or it would only ever use the default kubeconfig
settings.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -5,50 +5,41 @@ package e2e
 
 import (
 	"context"
-	"flag"
-	"fmt"
-	"os"
-	"os/user"
-	"path/filepath"
-	"strconv"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/open-cluster-management/governance-policy-framework/test/common"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 )
 
 var (
-	userNamespace          string
-	clusterNamespace       string
-	clientHub              kubernetes.Interface
-	clientHubDynamic       dynamic.Interface
-	clientManaged          kubernetes.Interface
-	clientManagedDynamic   dynamic.Interface
-	gvrPod                 schema.GroupVersionResource
-	gvrPolicy              schema.GroupVersionResource
-	gvrIamPolicy           schema.GroupVersionResource
-	gvrCertPolicy          schema.GroupVersionResource
-	gvrConfigurationPolicy schema.GroupVersionResource
-	gvrPlacementBinding    schema.GroupVersionResource
-	gvrPlacementRule       schema.GroupVersionResource
-	gvrRole                schema.GroupVersionResource
-	gvrCRD                 schema.GroupVersionResource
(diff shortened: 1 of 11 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -5,50 +5,41 @@ package e2e
...
+	clientHub             kubernetes.Interface
+	clientHubDynamic      dynamic.Interface
+	clientManaged         kubernetes.Interface
+	clientManagedDynamic  dynamic.Interface
+
+	gvrPod                 = schema.GroupVersionResource{Version: "v1", Resource: "pods"}
+	gvrNS                  = schema.GroupVersionResource{Version: "v1", Resource: "namespaces"}
+	gvrRole                = schema.GroupVersionResource{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "roles"}
+	gvrCRD                 = schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1beta1", Resource: "customresourcedefinitions"}
+	gvrPolicy              = schema.GroupVersionResource{Group: "policy.open-cluster-management.io", Version: "v1", Resource: "policies"}
```
mprahl, line 36: Optional: What do you think about putting these GVRs in the `common` package? I noticed this one is redefined in `policy-collection` and I think it'd be useful in general to have these for other test packages in the future.

test/e2e/e2e_suite_test.go
```diff
@@ -5,50 +5,41 @@ package e2e
...
-	gvrCRD                 schema.GroupVersionResource
-	gvrNS                  schema.GroupVersionResource
-	gvrK8sRequiredLabels   schema.GroupVersionResource
-	kubeconfigHub          string
-	kubeconfigManaged      string
-	defaultTimeoutSeconds  int
+	clusterNamespace = "managed"
 
-	defaultImageRegistry string
+	defaultTimeoutSeconds int
```
mprahl, line 25: Why is this variable defined here rather than use `common.defaultTimeoutSeconds` directly? I have the same question for `userNamespace`.

mprahl (commented):
This looks great! I just have a couple of small comments.

test/e2e/e2e_suite_test.go
JustinKuli: Yeah - the question becomes which ones to include in `common` (all? *any* that are shared?), and is it worth it to update all the times these appear to prepend `common.<foo>`.

In my opinion, it wasn't necessary yet, but with another set of tests maybe we'll find a better set of commonly used GVRs

mprahl: @JustinKuli I was thinking it might be nice to just have them all in a separate file in the `common` package? They don't necessarily need to be used by multiple tests yet, but it sets up a convention to reduce duplication in the future.

test/e2e/e2e_suite_test.go
JustinKuli: Good questions. First, these things appear a ton in the tests themselves. So although I could update all of those occurrences to prepend `common.`, I'd rather not - in addition to making this PR huge, it would make some lines incredibly long.

By being lazy and requiring `userNamespace = common.UserNamespace` in the BeforeSuite, it also provides an obvious point where a developer could interject their own custom value as needed. For example, the tests here specify "managed" instead of `common.ClusterNamespace`.

Devs could also use command line flags to set things like userNamespace, and it's helpful to have those flags defined in the common package. BUT, as a quirk of flag parsing and ginkgo, those flags can't be parsed in `init` functions, so they won't have the right values until the BeforeSuite function. That's why we can't for example do `var userNamespace = common.UserNamespace`.

mprahl: Thanks for the explanation!

test/e2e/e2e_suite_test.go
JustinKuli: That sounds reasonable, I'll do that!

mprahl (approved):

PR #139, opened 2021-08-18
Add a canary test for the IAM policy limitclusteradmin

mprahl:
This canary test is added in addition to the existing E2E tests in order to
test that RHBZ#1981127 is resolved. This must be done as a canary
test since it queries for OpenShift group memberships and this is not
supported on other Kubernetes distributions.

Related to:
https://github.com/open-cluster-management/backlog/issues/14166

### DIFF
--- a/test/policy-collection/policy_comp_operator_test.go
+++ b/test/policy-collection/policy_comp_operator_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/open-cluster-management/governance-policy-propagator/test/utils"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 )
 
 func isOCP46andAbove() bool {
@@ -80,6 +79,8 @@ var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] Test compliance opera
 	const compPolicyName = "policy-comp-operator"
 	const compE8ScanPolicyURL = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/main/stable/CM-Configuration-Management/policy-compliance-operator-e8-scan.yaml"
 	const compE8ScanPolicyName = "policy-e8-scan"
+	var getComplianceState func() interface{}
+
 	BeforeEach(func() {
 		if !isOCP46andAbove() {
 			Skip("Skipping as compliance operator is only supported on OCP 4.6 and above")
@@ -87,6 +88,9 @@ var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] Test compliance opera
 		if !canCreateOpenshiftNamespaces() {
 			Skip("Skipping as compliance operator requires the ability to create the openshift-compliance namespace")
 		}
+
+		// Assign this here to avoid using nil pointers as arguments
+		getComplianceState = common.GetComplianceState(clientHubDynamic, userNamespace, compPolicyName, clusterNamespace)
 	})
 	Describe("Clean up before all", func() {
 		It("clean up compliance scan e8", func() {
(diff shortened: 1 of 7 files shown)


test/resources/iam_policy/group.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Group
+metadata:
+  name: policy-limitclusteradmin
+users:
+  - rachel.green
+  - monica.geller
+  - phoebe.buffay
+  - joey.tribbiani
+  - chandler.bing
+  - ross.geller
```
JustinKuli, line 11: :joy:

test/policy-collection/policy_iam_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/open-cluster-management/governance-policy-framework/test/common"
+	policiesv1 "github.com/open-cluster-management/governance-policy-propagator/pkg/apis/policy/v1"
+	"github.com/open-cluster-management/governance-policy-propagator/test/utils"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+const iamPolicyName = "policy-limitclusteradmin"
+const iamPolicyURL = "https://raw.githubusercontent.com/open-cluster-management/policy-collection/main/stable/AC-Access-Control/" + iamPolicyName + ".yaml"
+
+func getComplianceState() interface{} {
```
JustinKuli, line 15: Can we make this take the name of the policy and namespace, and bring it into the common package? I'd be surprised if other tests can't use something like this.

JustinKuli: Or I guess it could go into propagator/test/utils, but I personally don't like that as a dependency. I'm never sure (without testing) how that dependency will get updated.

JustinKuli (commented):
Seems fine, just a couple comments

test/policy-collection/policy_iam_test.go
mprahl: Good idea. I'll make this a separate commit.

mprahl: @JustinKuli is the first commit what you had in mind?

test/policy-collection/policy_comp_operator_test.go
```diff
@@ -80,13 +79,18 @@ var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] Test compliance opera
...
 	BeforeEach(func() {
 		if !isOCP46andAbove() {
 			Skip("Skipping as compliance operator is only supported on OCP 4.6 and above")
 		}
 		if !canCreateOpenshiftNamespaces() {
 			Skip("Skipping as compliance operator requires the ability to create the openshift-compliance namespace")
 		}
+
+		// Assign this here to avoid using nil pointers as arguments
+		getComplianceState = common.GetComplianceState(clientHubDynamic, userNamespace, compPolicyName, clusterNamespace)
```
JustinKuli, line 93: Interesting - I wasn't expecting this. What would be a nil pointer if you don't do this?

test/policy-collection/policy_gatekeeper_operator_downstream_test.go
```diff
@@ -14,14 +14,20 @@ import (
...
+
 	BeforeEach(func() {
 		if isOCP44() {
 			Skip("Skipping as this is ocp 4.4")
 		}
+
+		// Assign this here to avoid using nil pointers as arguments
+		getComplianceState = func(policyName string) func() interface{} {
+			return common.GetComplianceState(clientHubDynamic, userNamespace, policyName, clusterNamespace)
+		}
```
JustinKuli, line 30: Ok, in this case it's a little different because you're adding in some of the values you'd otherwise need to repeat... which is cool.

JustinKuli (approved):
Ok - I see. I was expecting to just use something like
```
Eventually(common.GetComplianceState(clientHubDynamic, userNamespace, "my-policy", clusterNamespace), defaultTimeoutSeconds*2, 1).Should(Equal(policiesv1.NonCompliant))
```
every time... and since the first case I saw in review was the compliance operator where the policy name never changes, I got the wrong idea of what you were doing.

Defining `getComplianceState` and saving the values that you'd otherwise need to repeat a lot (eg in the gatekeeper tests) is even better than what I was originally thinking. Nice!

test/policy-collection/policy_comp_operator_test.go
JustinKuli: `clientHubDynamic` et al would be nil if you tried to define this outside of the suite, is that right? Doesn't matter much, just curious.

mprahl: For whatever reason, `userNamespace` ended up being a nil pointer. I think it's the timing issue you mentioned in #137 about the parsing of CLI arguments by ginkgo.

PR #142, opened 2021-08-19
Add metrics tests to run on canaries

JustinKuli:
The tests in the propagator itself verifies some of these same things,
but the tests here use the service and rbac-proxy set up in the Helm
chart, to verify they are working as expected. One additional test is
to verify that the metrics can't be seen by unauthorized users.

This utilizes the `oc` cli's ability to expose services as a route, so
that we can curl directly to the metrics endpoint. One alternative that
I considered was using a port-forward, but that was difficult to manage.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/14992

### DIFF
--- a/test/integration/integration_suite_test.go
+++ b/test/integration/integration_suite_test.go
@@ -61,13 +61,13 @@ var _ = BeforeSuite(func() {
 	}
 
 	By("Create Namespace if needed")
-	namespaces := clientHub.CoreV1().Namespaces()
-	if _, err := namespaces.Get(context.TODO(), userNamespace, metav1.GetOptions{}); err != nil && errors.IsNotFound(err) {
-		Expect(namespaces.Create(context.TODO(), &corev1.Namespace{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: userNamespace,
-			},
-		}, metav1.CreateOptions{})).NotTo(BeNil())
+	_, err := clientHub.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: userNamespace,
+		},
+	}, metav1.CreateOptions{})
+	if err != nil {
+		Expect(errors.IsAlreadyExists(err)).Should(BeTrue())
 	}
-	Expect(namespaces.Get(context.TODO(), userNamespace, metav1.GetOptions{})).NotTo(BeNil())
+	Expect(clientHub.CoreV1().Namespaces().Get(context.TODO(), userNamespace, metav1.GetOptions{})).NotTo(BeNil())
 })
--- a/test/integration/integration_suite_test.go
+++ b/test/integration/integration_suite_test.go
@@ -1,0 +1,73 @@
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
+
+	"github.com/open-cluster-management/governance-policy-framework/test/common"
+)
+
+var (
+	userNamespace         string
+	clusterNamespace      string
+	kubeconfigHub         string
+	kubeconfigManaged     string
+	defaultTimeoutSeconds int
+	clientHub             kubernetes.Interface
+	clientHubDynamic      dynamic.Interface
+	clientManaged         kubernetes.Interface
+	clientManagedDynamic  dynamic.Interface
+	getComplianceState    func(policyName string) func() interface{}
+)
+
+func TestIntegration(t *testing.T) {
+	RegisterFailHandler(Fail)
+	junitReporter := reporters.NewJUnitReporter("../../test-output/integration.xml")
+	RunSpecsWithDefaultAndCustomReporters(t, "Policy Framework repo integration Suite", []Reporter{junitReporter})
+}
+
+func init() {
+	klog.SetOutput(GinkgoWriter)
+	klog.InitFlags(nil)
+}
+
+var _ = BeforeSuite(func() {
+	By("Setup hub and managed client")
+	kubeconfigHub = common.KubeconfigHub
+	kubeconfigManaged = common.KubeconfigManaged
+	userNamespace = common.UserNamespace
+	clusterNamespace = common.ClusterNamespace
+	defaultTimeoutSeconds = common.DefaultTimeoutSeconds
+
+	clientHub = common.NewKubeClient("", kubeconfigHub, "")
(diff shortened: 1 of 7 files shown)


test/integration/integration_suite_test.go
```diff
@@ -0,0 +1,73 @@
...
+	clientManaged = common.NewKubeClient("", kubeconfigManaged, "")
+	clientManagedDynamic = common.NewKubeClientDynamic("", kubeconfigManaged, "")
+
+	getComplianceState = func(policyName string) func() interface{} {
+		return common.GetComplianceState(clientHubDynamic, userNamespace, policyName, clusterNamespace)
+	}
+
+	By("Create Namespace if needed")
+	namespaces := clientHub.CoreV1().Namespaces()
+	if _, err := namespaces.Get(context.TODO(), userNamespace, metav1.GetOptions{}); err != nil && errors.IsNotFound(err) {
```
mprahl, line 65: Optional: You could just try creating the namespace and then ignore errors saying it already exists. This reduces an API call:

For example:
```go
		if err != nil {
			Expect(errors.IsAlreadyExists(err)).Should(BeTrue())
		}
```

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+			routeList, err = clientHubDynamic.Resource(common.GvrRoute).Namespace(ocmNS).List(context.TODO(), metav1.ListOptions{LabelSelector: propagatorMetricsSelector})
+			if err != nil {
+				return err
+			}
+			return len(routeList.Items)
+		}, defaultTimeoutSeconds, 1).Should(Or(Equal(0), Equal(1)))
+
+		if len(routeList.Items) == 0 {
+			By("Exposing the metrics service as a route")
+			_, err = oc("expose", "service", metricsSvc.Name, "-n", ocmNS, `--overrides={"spec":{"tls":{"termination":"reencrypt"}}}`)
```
mprahl, line 51: That's a nice little shortcut compared to creating the route by hand!

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+	if len(args) > 0 && args[0] != "whoami" {
+		fmt.Println(string(output))
+	}
+	return string(output), err
+}
+
+func curlMetricsRoute(args ...string) (string, error) {
+	argList := []string{"-k", "https://" + routeURL + "/metrics"}
+	argList = append(argList, args...)
+	output, err := exec.Command("curl", argList...).CombinedOutput()
```
mprahl, line 147: Could you please use [net/http](https://pkg.go.dev/net/http) instead? That's OS agnostic, you don't have to parse stdout/stderr, and it feels a lot cleaner.

Also, don't forget to add a timeout since we don't want the HTTP request to wait forever if that server is acting funky.

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+		By("Got the metrics route url: " + routeURL)
+	})
+	It("Checks that the endpoint does not expose metrics without auth", func() {
+		Eventually(func() interface{} {
+			resp, err := curlMetricsRoute()
+			if err != nil {
+				return err
+			}
+			return resp
+		}, defaultTimeoutSeconds, 1).Should(ContainSubstring("Unauthorized"))
```
mprahl, line 74: After you follow my comment on using net/http instead of executing curl, could you please check the HTTP status code directly instead?

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Eventually(
+			getComplianceState(compliantPolicyName),
+			defaultTimeoutSeconds,
+			1,
+		).Should(Equal(policiesv1.Compliant))
+
+		By("Checking the policy metric")
+		token, err := oc("whoami", "-t")
+		Expect(err).To(BeNil())
+		regex := `(?m)` + metricName + `{.*policy="` + compliantPolicyName + `.*} 0$`
```
mprahl, line 99: Could you please add a comment above this of the string you are capturing? This make this easier to read.

mprahl:
```suggestion
		regex := `(?m)` + metricName + `{.*policy="` + compliantPolicyName + `".*} 0$`
```

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+		Eventually(
+			getComplianceState(noncompliantPolicyName),
+			defaultTimeoutSeconds,
+			1,
+		).Should(Equal(policiesv1.NonCompliant))
+
+		By("Checking the policy metric")
+		token, err := oc("whoami", "-t")
+		Expect(err).To(BeNil())
+		regex := `(?m)` + metricName + `{.*policy="` + noncompliantPolicyName + `.*} 1$`
```
mprahl, line 120: Since this is so similar to the previous regex, it might be nice to have a function that determines if it's a violation:
```go
func isPolicyNonCompliant(policyName, metricsOutput string) (bool, error) {
	regex := regexp.MustCompile(`(?m)` + metricName + `{.*policy="` + policyName + `".*} (?P<violation>\d+)$`)
	matches := regex.FindStringSubmatch(metricsOutput)
	if len(matches) == 0 {
		return false, fmt.Errorf("the policy %s was not present in the metrics output", policyName)
	}

	return matches[regex.SubexpIndex("violation")] == "1", nil
}
```

test/integration/policy_info_metric_test.go
JustinKuli: That's a good idea

test/integration/policy_info_metric_test.go
JustinKuli: Good point

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,185 @@
...
+				InsecureSkipVerify: true,
+			},
+		},
+		Timeout: 5 * time.Second,
+	}
+	req, err := http.NewRequest("GET", "https://"+routeURL+"/metrics", nil)
+	if err != nil {
+		return "", "", err
+	}
+	if len(authToken) > 0 {
```
mprahl, line 167: Optional: I find this more intuitive to read.
```suggestion
	if authToken != "" {
```

test/integration/policy_info_metric_test.go
```diff
@@ -0,0 +1,185 @@
...
+	propagatorMetricsSelector = "component=ocm-policy-propagator"
+	ocmNS                     = "open-cluster-management"
+	metricName                = "policy_governance_info"
+	compliantPolicyYaml       = "../resources/policy_info_metric/compliant.yaml"
+	compliantPolicyName       = "policy-metric-compliant"
+	noncompliantPolicyYaml    = "../resources/policy_info_metric/noncompliant.yaml"
+	noncompliantPolicyName    = "policy-metric-noncompliant"
+)
+
+var routeURL string
```
mprahl, line 34: Technically this isn't a URL since a URL would contain the protocol (i.e. `https`). So maybe call this `routeHost`?

mprahl (approved):
/hold

mprahl:
> /hold

Just unhold it if you want to merge it as is. If you do address the nitpicks, let me know and I'll approve it.

JustinKuli:
Good nits :)

/unhold

mprahl (approved):

PR #143, opened 2021-08-20
Use ioutil.ReadAll instead of io.ReadAll

JustinKuli:
The `io.ReadAll` function was added in Go version 1.16, so depending
on the version used by test runners, the tests might fail to compile.
The `ioutil.ReadAll` function still exists and can be used for
compatability.

Since Go 1.17 just came out, we will need to bump up versions in our go.mod and test runners (we currently use 1.14), but that is a separate issue. This will fix the e2e tests with minimal changes.

### DIFF
--- a/test/integration/policy_info_metric_test.go
+++ b/test/integration/policy_info_metric_test.go
@@ -6,7 +6,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"io"
+	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


JustinKuli:
Separate from the Travis config and the go.mod file, the e2etest Dockerfile looks like it uses Go v 1.15.

mprahl (approved):

dhaiducek (approved):

PR #144, opened 2021-08-23
Disable new metrics test

JustinKuli:
It is failing on the e2e clusters, partly because some of the tests are
not using the right kubeconfig, but also because the e2e clusters do not
use the updated deployment with the rbac proxy and service endpoint that
this suite would test.

### DIFF
--- a/build/run-test-image.sh
+++ b/build/run-test-image.sh
@@ -5,4 +5,3 @@
 set -e
 
 ginkgo -v --slowSpecThreshold=10 test/policy-collection -- -cluster_namespace=$MANAGED_CLUSTER_NAME
-ginkgo -v --slowSpecThreshold=10 test/integration -- -cluster_namespace=$MANAGED_CLUSTER_NAME


build/run-test-image.sh
```diff
@@ -5,4 +5,4 @@
 set -e
 
 ginkgo -v --slowSpecThreshold=10 test/policy-collection -- -cluster_namespace=$MANAGED_CLUSTER_NAME
-ginkgo -v --slowSpecThreshold=10 test/integration -- -cluster_namespace=$MANAGED_CLUSTER_NAME
+# ginkgo -v --slowSpecThreshold=10 test/integration -- -cluster_namespace=$MANAGED_CLUSTER_NAME
```
mprahl, line 8: Could you please delete the line instead? When we want to reenable it, we can always look at the git history to find the command. It's just best practice to not leave commented out code.
```suggestion
```

mprahl (approved):

PR #152, opened 2021-08-31
Handle exiterror

willkutler:
https://github.com/open-cluster-management/backlog/issues/15718


dhaiducek (approved):
LGTM. I'm not 100% sure how `CombinedOutput` and `Output` are different, but this certainly seems right...

PR #153, opened 2021-09-01
Handle exiterror

willkutler:
try change token based on new failures related to oc whoami

https://github.com/open-cluster-management/backlog/issues/15774

### DIFF
--- a/test/integration/policy_info_metric_test.go
+++ b/test/integration/policy_info_metric_test.go
@@ -4,8 +4,8 @@ package integration
 
 import (
 	"context"
-	"strings"
 	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


test/integration/policy_info_metric_test.go
```diff
@@ -18,6 +19,9 @@ const (
 	propagatorMetricsSelector = "component=ocm-policy-propagator"
 	ocmNS                     = "open-cluster-management"
 	metricName                = "policy_governance_info"
+	saName					  = "grc-framework-sa"
+	saNamespace				  = "kube-system"
```
dhaiducek, line 23: We shouldn't be messing with the `kube-system` namespace in our tests. Is there another namespace (new or one created for our tests) that we could use? Maybe just reuse `userNamespace`?

test/integration/policy_info_metric_test.go
```diff
@@ -78,8 +82,16 @@ var _ = Describe("Test policy_governance_info metric", func() {
 		common.OcHub("apply", "-f", compliantPolicyYaml, "-n", userNamespace)
 		// Don't need to check compliance - just need to guarantee there is a policy in the cluster
 
-		token, err := common.OcHub("whoami", "-t")
+		//get token from SA
+		_, err := common.OcHub("create", "serviceaccount", saName, "-n", saNamespace)
```
dhaiducek, line 86: We probably need to make sure we're cleaning up this ServiceAccount in "Cleans up", also.

dhaiducek (changes requested):
Good thinking on leveraging a ServiceAccount! There are other instances of the `oc whoami -t`, so those will need this token also:
- Checks that a compliant policy reports a metric of 0
- Checks that a noncompliant policy reports a metric of 1

dhaiducek (approved):
LGTM!

PR #169, opened 2021-09-22
Update propagator resource paths

JustinKuli:
Previously, we had copied resource and CRD yamls into this repo, and
were (mostly) using those for deployment. But those resources can drift,
and it will be better to reference the original locations. With the
changes in the propagator deployment, this is a good time to get it
started.

Where the copied resources are no longer required, I deleted them and
added a README pointing to better locations for them.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/15548


JustinKuli:
Since this uses updated paths from this PR, we'll need to wait for it to be merged: https://github.com/open-cluster-management/governance-policy-propagator/pull/99

JustinKuli:
Re-running the travis build now that the propagator PR has been merged, then I think this is ready for review / merge

JustinKuli:
@mprahl when we discussed this you brought up that applying the CRDs from raw.githubusercontent addresses could run into rate-limiting... I had a bash script send over 5000 requests to that endpoint and didn't have any issues, so I'm hoping we will be ok here. From the github docs it sounds like only API requests would be rate-limited, for some reason.

mprahl (approved):

PR #183, opened 2021-10-04
Migrate image builds to prow

dhaiducek:


ycao56:
image
This is not needed as we don't run Travis build for PR.

dhaiducek:
/unhold All set! The branch protection bot doesn't run very frequently--maybe every 6 hours or so, so I was waiting to see those rules update in the Settings, but it looks like they updated here also. Here goes nothing! 😄 

image

PR #212, opened 2021-11-02
Use the IamPolicy CRD directly from the repo

mprahl:
This reduces the chance of the two versions drifting from each other.


ycao56 (approved):

JustinKuli (approved):

mprahl:
I assume the CI failure is due to it not using a snapshot that includes the operator-sdk upgrade of the IAM policy controller. I'll retry this tomorrow.

mprahl:
> I assume the CI failure is due to the latest tag not yet including the operator-sdk upgrade of the IAM policy controller. I'll retry this tomorrow.

It ended up being an issue where the controller was trying to find the hub config from the wrong secret. My latest push patches the container args to include `"--hubconfig-secret-name=hub-kubeconfig"`.

JustinKuli (approved):
/lgtm

I think there needs to be a discussion on how we handle the lease (and hubconfig secret it uses) in general, but it doesn't need to block this PR.

PR #214, opened 2021-11-10
Send travis notifications to a webhook

JustinKuli:
In order to collect more test summary information, we might want to use
a webhook, as opposed to relying on messages sent through slack. At the
very least, I want to know what information is exactly in the webhook
that we might be able to use.

The endpoint is encrypted, so it doesn't get flooded or anything.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/17431

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,4 +120,6 @@ notifications:
       - "Build details: %{build_url}"
     on_success: always
     on_failure: always
-    
+  webhooks:
+    urls:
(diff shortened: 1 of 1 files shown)


JustinKuli:
/hold

I can probably get the information I need without merging this

ycao56 (approved):

JustinKuli:
/unhold

We were considering changing the PR to get the webhook notifications just on the PR run, but since it's currently from a fork that wouldn't work. And I'd need to time the push so the PR tests didn't interfere with scheduled tests. So this will be fine, and potentially get more data points.

PR #227, opened 2021-12-15
Periodic job that can be run to detect prow publish failures

gparvin:
There's a desire to know when prow jobs fail on our release branches
since publishes to those branches happens automatically and failures
generally do not grab our attention.  This script can help and could
send slack messages through the prow framework which only allows
slack messages on periodic jobs -- not our existing post submit jobs.

Refs:
 - https://github.com/open-cluster-management/backlog/issues/17925

### DIFF
--- a/build/periodic.sh
+++ b/build/periodic.sh
@@ -1,0 +1,123 @@
+#!/bin/sh
+
+# Validate the pipeline is up to date and that no failed prow jobs exist
+
+COMPONENT_ORG=stolostron
+CHECK_RELEASES="2.3 2.4 2.5"
+
+# Clone the repositories needed for this script to work
+cloneRepos() {
+	if [ ! -d "pipeline" ]; then
+		git clone https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/$COMPONENT_ORG/policy-grc-squad.git
+	fi
+	# Only clone pipeline if it doesn't already exist.
+	if [ ! -d "pipeline" ]; then
+		git clone https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/$COMPONENT_ORG/pipeline.git
+	fi
+	REPOS=$(cat policy-grc-squad/main-branch-sync/repo.txt | grep policy | grep -v framework)
+	for repo in $REPOS; do
+		printf '%s\n' "Updating $repo ...."
+		git clone git@github.com:$repo.git $repo
+	done
+}
+
+# return the most recent git sha for a repository's release branch
+getGitSha() {
+	component=$1
+	release=release-$2
+	cd $COMPONENT_ORG/$component
+	co=`git checkout $release`
+	GITSHA=`git log -n 1 --no-decorate --pretty=oneline | awk '{print $1}'`
+	echo "$GITSHA"
+	cd $BASEDIR
+}
+
+getPipelineValue() {
+	component="$1"
+	release="${2}-integration"
+	key="$3"
+
+	cd pipeline
+	co=$(git checkout "$release")
+	value=`jq '.[] |select(.["image-name"] == "'$component'") | .["'$key'"]' manifest.json | sed 's/"//g'`
+	echo "$value"
+	cd $BASEDIR
+}
+
+checkProwJob() {
+	component="$1"
+	release="$2"
+
+	rcode=0
+	# This is a hack to get data from the openshift ci prow
+	# sample curl: curl -H "content-type: application/xml" https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/branch-ci-open-cluster-management-iam-policy-controller-release-2.5-publish
+        # which contains this:   var allBuilds = [{"SpyglassLink":"/view/gs/origin-ci-test/logs/branch-ci-open-cluster-management-iam-policy-controller-release-2.5-publish/1468625788839399424","ID":"1468625788839399424","Started":"2021-12-08T16:57:31Z","Duration":103000000000,"Result":"FAILURE","Refs":{"org":"open-cluster-management","repo":"iam-policy-controller","repo_link":"https://github.com/open-cluster-management/iam-policy-controller","base_ref":"release-2.5","base_sha":"567b3597e8324a4c56ec8f1d717ae15d9671e4a8","base_link":"https://github.com/open-cluster-management/iam-policy-controller/compare/d544db4214b4...567b3597e832"}}];
+	echo "Checking prow jobs for a failure with component $component."
+	jobs="publish images latest-image-mirror"
+	for job in $jobs; do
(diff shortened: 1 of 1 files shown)


gparvin:
The idea is a periodic prow job could call this... in particular I'm expecting it to be our framework component.

Thinking about how to handle the releases list which is currently hard coded.

dhaiducek:
I like this idea since we already post to Slack through this Travis job. Taking a look now.

build/periodic.sh
```diff
@@ -0,0 +1,111 @@
...
+checkProwJob() {
+	component="$1"
+	release="$2"
+
+	rc=0
+	# This is a hack to get data from the openshift ci prow
+	# sample curl: curl -H "content-type: application/xml" https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/branch-ci-open-cluster-management-iam-policy-controller-release-2.5-publish
+        # which contains this:   var allBuilds = [{"SpyglassLink":"/view/gs/origin-ci-test/logs/branch-ci-open-cluster-management-iam-policy-controller-release-2.5-publish/1468625788839399424","ID":"1468625788839399424","Started":"2021-12-08T16:57:31Z","Duration":103000000000,"Result":"FAILURE","Refs":{"org":"open-cluster-management","repo":"iam-policy-controller","repo_link":"https://github.com/open-cluster-management/iam-policy-controller","base_ref":"release-2.5","base_sha":"567b3597e8324a4c56ec8f1d717ae15d9671e4a8","base_link":"https://github.com/open-cluster-management/iam-policy-controller/compare/d544db4214b4...567b3597e832"}}];
+	echo "Checking prow jobs for a failure with component $component."
+	OUTPUT=$(curl -s https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/branch-ci-${COMPONENT_ORG}-${component}-release-${release}-publish)
```
dhaiducek, line 52: I think we should also have the `latest-image-mirror` and `images` jobs in here for completeness.

build/periodic.sh
```diff
@@ -0,0 +1,111 @@
...
+	JSON=$(echo "$OUTPUT" | grep "var allBuilds" | sed 's/  var allBuilds =//' | sed 's/;$//' | jq '.[0]')
```
dhaiducek, line 53: Holy moly--great find on the `allBuilds` variable!

And I see this only checks the most recent job. I was thinking of getting statuses for everything but I like this approach better since if one fails but the next one succeeds, that should correct any error.

dhaiducek (commented):
Thanks for the PR! I like this approach. It's a bit hack-y, but it looks like that's what we need right now (at least until a `reporter_config` is supported for generated configs).

build/periodic.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/sh
+
+# Validate the pipeline is up to date and that no failed prow jobs exist
+
+COMPONENT_ORG=open-cluster-management
```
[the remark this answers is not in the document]
gparvin: It doesn't work with the new org yet.  So I still have the old one in there.

build/periodic.sh
```diff
@@ -0,0 +1,111 @@
...
+CHECK_RELEASES="2.3 2.4 2.5"
+
+# Clone the repositories needed for this script to work
+cloneRepos() {
+	git clone https://github.com/$COMPONENT_ORG/policy-grc-squad.git
+	git clone https://github.com/$COMPONENT_ORG/pipeline.git
```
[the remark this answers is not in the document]
gparvin: For now I took this out and just have the repos listed in a variable.  Let me know what you think.

build/periodic.sh
```diff
@@ -0,0 +1,111 @@
...
+			echo "****"
+			echo "WARNING!!!  SHA mismatch in pipeline and $repo $RELEASE repositories."
+        		echo "   pipeline: $pipelinesha"
+        		echo "   $repo: $gitsha"
+			echo "***"
+        		rc=1
+		fi
+
+		# make sure the docker image is available
+		docker pull quay.io/$COMPONENT_ORG/$repo:$imagetag
```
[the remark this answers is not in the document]
gparvin: Great idea, I am using your idea now.

build/periodic.sh
gparvin: I added these.  Please take a look since I just turned it into a loop and ran the code again.  It seems to work.

build/periodic.sh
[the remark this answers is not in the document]
dhaiducek: `pipeline` is also private, unfortunately.

build/periodic.sh
dhaiducek: Nice--looks good!

build/periodic.sh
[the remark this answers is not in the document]
gparvin: @dhaiducek If this is called from a properly authenticated prow job, could that work?  It seems like if they can run the publish that updates pipeline, I should be able to get something similar to happen for this.

build/periodic.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/sh
+
+# Validate the pipeline is up to date and that no failed prow jobs exist
+
+COMPONENT_ORG=stolostron
+#COMPONENT_ORG=open-cluster-management
```
dhaiducek, line 6: Clean up commented line

build/periodic.sh
```diff
@@ -0,0 +1,130 @@
...
+$COMPONENT_ORG/config-policy-controller
+$COMPONENT_ORG/governance-policy-propagator
+$COMPONENT_ORG/governance-policy-spec-sync
+$COMPONENT_ORG/governance-policy-status-sync
+$COMPONENT_ORG/governance-policy-template-sync
+$COMPONENT_ORG/iam-policy-controller"
+
+# Clone the repositories needed for this script to work
+cloneRepos() {
+	#git clone https://github.com/$COMPONENT_ORG/policy-grc-squad.git
```
dhaiducek, line 19: Should we restore this line?

dhaiducek (changes requested):

build/periodic.sh
gparvin: Done

build/periodic.sh
gparvin: I restored it so the repo list is calculated now instead of hard coded

build/periodic.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/sh
+
+# Validate the pipeline is up to date and that no failed prow jobs exist
+
+COMPONENT_ORG=stolostron
+CHECK_RELEASES="2.3 2.4 2.5"
```
dhaiducek, line 6: For a future enhancement, we could use https://github.com/stolostron/release/blob/master/CURRENT_RELEASE to get the current release and create these versions dynamically.

dhaiducek (approved):
LET'S DO THIS!!!

gparvin:
I fixed the issue with BASEDIR

dhaiducek (approved):

PR #233, opened 2022-01-04
Update org references to `stolostron`

ycao56:
Make e2e test to run again. More changes to come after GO module name and API path is updated in dependent repos
https://github.com/stolostron/backlog/issues/18738


dhaiducek (approved):
Some good finds in there! LGTM!

PR #251, opened 2022-01-20
There's a problem with doing the git clones from prow

gparvin:
Switch one of the git clones from using ssh to https.  The ssh style
of doing a clone doesn't work in prow.

### DIFF
--- a/build/periodic.sh
+++ b/build/periodic.sh
@@ -17,7 +17,7 @@ cloneRepos() {
 	REPOS=$(cat policy-grc-squad/main-branch-sync/repo.txt | grep policy | grep -v framework)
 	for repo in $REPOS; do
 		printf '%s\n' "Updating $repo ...."
-		git clone git@github.com:$repo.git $repo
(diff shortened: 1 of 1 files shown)


build/periodic.sh
```diff
@@ -17,7 +17,7 @@ cloneRepos() {
 	REPOS=$(cat policy-grc-squad/main-branch-sync/repo.txt | grep policy | grep -v framework)
 	for repo in $REPOS; do
 		printf '%s\n' "Updating $repo ...."
-		git clone git@github.com:$repo.git $repo
+		git clone https://${GITHUB_USER}:${GITHUB_TOKEN}@github.com/$repo.git $repo
```
dhaiducek, line 20: Leaving the user and token off here will work since all the repos are public.

dhaiducek (changes requested):

build/periodic.sh
gparvin: I made the requested update. Thanks!

dhaiducek (approved):

PR #252, opened 2022-01-25
Prow may match our errors better using ERROR instead of WARNING

gparvin:
Enhance the experience so error message "may" appear in the prow log
summary without having to drill into the raw log details.  Dale
saw the prow error checking looks for `ERROR` so using that value
instead of `WARNING` in the message.

Refs:
 - https://github.com/stolostron/backlog/issues/17925


dhaiducek (approved):

ycao56 (approved):

ycao56:
Force merging it as the failure is not related to this PR. Fixing the failure in a different PR.

PR #266, opened 2022-02-21
Add additional checks to Prow status script

dhaiducek:
The `governance-policy-addon-controller` repo was private (it's since become public) and also doesn't have `2.3` or `2.4` images. To address these "false positives", this PR adds:
- An additional check for error messages when querying Quay
- An additional check when an image wasn't found in Pipeline


gparvin (approved):
Thanks for the fix!

PR #272, opened 2022-03-02
Move over clean up script

ycao56:

### DIFF
--- a/build/clean-up-cluster.sh
+++ b/build/clean-up-cluster.sh
@@ -3,20 +3,58 @@
 
 
 set -e
-UI_CURRENT_IMAGE=$1
 
-make docker/login
-export DOCKER_URI=quay.io/stolostron/grc-ui-tests:latest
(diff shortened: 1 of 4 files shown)


build/clean-up-cluster.sh
```diff
@@ -3,20 +3,57 @@
...
 
 set -e
-UI_CURRENT_IMAGE=$1
 
-make docker/login
-export DOCKER_URI=quay.io/stolostron/grc-ui-tests:latest
-make docker/pull
+function hub() {
+    echo "Hub: clean up"
+    oc delete ns duplicatetest || true
```
mprahl, line 9: Instead of `|| true`, could you use the `--ignore-not-found` flag instead?

ycao56: Great. Didn't know the flag exist. Will update

JustinKuli (approved):

PR #276, opened 2022-03-03
Add metadata to BeforeSuite and AfterSuite phases

dhaiducek:
Not the cleanest, but I think it's better than removing the XML nodes entirely.

Addresses:
- https://github.com/stolostron/backlog/issues/20431


ycao56 (approved):

dhaiducek:
Verified JUnit looks okay. Reran test to see if the failed metrics test passes.

PR #286, opened 2022-03-13
Patch enforcementAction to deny

ycao56:
Removing enforcementAction does not always work as the policy is set to 
`musthave`. Explicitly set enforcementAction to deny in the policy to 
ensure the gatekeeper constraint is set to deny to avoid test flakiness.
Also patch the enforcementAction before placing the policy.


dhaiducek (approved):
LGTM!

PR #287, opened 2022-03-14
Add a 5 second timeout to oc login

mprahl:
This should resolve the failure in BVT 247795519.


ycao56 (approved):

test/common/user_management.go
```diff
@@ -54,7 +55,11 @@ func GetKubeConfig(server, username, password string) (string, error) {
 		return "", fmt.Errorf("failed to close the temporary kubeconfig")
 	}
 
-	output, err := exec.Command(
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
JustinKuli, line 58: Is 5 seconds enough? I feel like I occasionally get noticeable slowdown on some `oc` commands, if the api server is busy (which it might be, in the middle of automated tests from other teams). So I might be more comfortable with 15 or 20 seconds here? That would be about the point on my computer where I might `ctrl-c` it and retry

JustinKuli (changes requested):

test/common/user_management.go
mprahl: My thought was that since `Eventually` will keep calling it, it'd be fine. However, I increased it to 20 seconds since it doesn't hurt.

PR #290, opened 2022-03-15
Create a test for the acm hardening policy set

gparvin:
The acm hardening policy set has been moved to a stable directory.
This is a test that will help make sure it continues to have some
validation.

Refs:
 - https://github.com/stolostron/backlog/issues/20337

### DIFF
--- a/test/policy-collection/policy_generator_acm_hardening.go
+++ b/test/policy-collection/policy_generator_acm_hardening.go
@@ -1,9 +1,10 @@
 // Copyright Contributors to the Open Cluster Management project
 
-package integration
+package e2e
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 
@@ -27,9 +28,6 @@ func cleanup(namespace string, secret string, user common.OCPUser) {
 		Expect(err).Should(BeNil())
 	}
 
-	err = common.CleanupOCPUser(clientHub, clientHubDynamic, secret, user)
-	Expect(err).Should(BeNil())
-
 	// Wait for the namespace to be fully deleted before proceeding.
 	Eventually(
 		func() bool {
@@ -41,11 +39,20 @@ func cleanup(namespace string, secret string, user common.OCPUser) {
 		defaultTimeoutSeconds,
 		1,
 	).Should(BeTrue())
+
+	err = common.CleanupOCPUser(clientHub, clientHubDynamic, secret, user)
+	Expect(err).Should(BeNil())
+
+	err = clientHub.CoreV1().Secrets("openshift-config").Delete(context.TODO(), secret, metav1.DeleteOptions{})
+	if !k8serrors.IsNotFound(err) {
+		Expect(err).Should(BeNil())
+	}
 }
 
-var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the Policy Generator in an App subscription", func() {
-	const namespace = "grc-e2e-policy-generator"
+var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the ACM Hardening generated PolicySet in an App subscription", func() {
+	const namespace = "policies"
 	const secret = "grc-e2e-subscription-admin-user"
+	const clustersetRoleName = "grc-e2e-clusterset-role"
 	const subAdminBinding = "open-cluster-management:subscription-admin"
 	ocpUser := common.OCPUser{
 		ClusterRoles: []types.NamespacedName{
@@ -54,6 +61,7 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the Policy Generator in an Ap
 				Name:      "admin",
 				Namespace: namespace,
 			},
+			{Name: clustersetRoleName},
 		},
 		// To be considered a subscription-admin you must be part of this cluster role binding.
 		// Having the proper role in another cluster role binding does not work.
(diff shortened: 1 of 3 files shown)


test/policy-collection/policy_generator_acm_hardening.go
```diff
@@ -0,0 +1,289 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+// generateInsecurePassword is a random password generator from 15-30 bytes. It is insecure
+// since the characters are limited to just hex values (i.e. 1-9,a-f) from the random bytes. An
+// error is returned if the random bytes cannot be read.
+func generateInsecurePassword() (string, error) {
```
mprahl, line 28: Could you please make this a public function in `test/common` to avoid duplicating this from the other Policy Generator test?

test/policy-collection/policy_generator_acm_hardening.go
```diff
@@ -0,0 +1,289 @@
...
+					compliant, found, myerr := unstructured.NestedString(policy.Object, "status", "compliant")
+					if myerr != nil {
+						return myerr
+					}
+					if !found {
+						return fmt.Errorf("failed to find the compliant field of the policy status")
+					} else {
+						if compliant != "Compliant" {
+							return fmt.Errorf("The policy is not compliant")
+						}
```
mprahl, line 246:
```suggestion
					} else if compliant != "Compliant" {
						return fmt.Errorf("The policy is not compliant")
```

mprahl (changes requested):
This looks good, but I noticed that two of the four policies are not checked in the test. Is this intentional?

test/resources/policy_generator/acm-hardening_subscription.yaml
```diff
@@ -0,0 +1,77 @@
...
+metadata:
+  annotations:
+    apps.open-cluster-management.io/git-branch: main
+    apps.open-cluster-management.io/git-path: policygenerator/policy-sets/stable/acm-hardening
+    apps.open-cluster-management.io/reconcile-option: merge
+  labels:
+    app: acm-hardening
+  name: acm-hardening-subscription
+spec:
+  allow:
```
ycao56, line 46: allow list is no longer needed

test/resources/policy_generator/acm-hardening_subscription.yaml
```diff
@@ -0,0 +1,77 @@
+---
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: ManagedClusterSetBinding
```
ycao56, line 3: is this needed? Not sure what it is for. The Application is using placementrule so this is not used. The policies are created in `policies ` ns. So if this is for policies, it should be in `policies` ns. Did I understand it correctly?

mprahl: You're right. It should be for the `policies` namespace.

gparvin: It is a bit confusing I think.  The Subscription uses `PlacementRule` but the `PolicySet` uses `Placement`.  So this resource is needed by the `PolicySet` which is deployed by the subscription.

mprahl: I see. The `kubectl apply` specifies `-n namespace` which is `-n policies`.

test/policy-collection/policy_generator_acm_hardening.go
gparvin: yes, thanks for this comment.  I have moved the function.

test/policy-collection/policy_generator_acm_hardening.go
gparvin: completed this one too

test/resources/policy_generator/acm-hardening_subscription.yaml
gparvin: I took out the allow list and the tests are working fine.  Thanks!

ycao56 (dismissed):

test/policy-collection/policy_generator_acm_hardening.go
```diff
@@ -0,0 +1,274 @@
...
+
+		_, err = clientHub.RbacV1().ClusterRoles().Create(
+			context.TODO(), &clusterSetRule, metav1.CreateOptions{},
+		)
+		if err != nil {
+			Expect(k8serrors.IsAlreadyExists(err)).Should(BeTrue())
+		}
+
+		By("Verifying that the managed clusterset namespaced binding exists")
+		clustersetBinding := rbacv1.ClusterRoleBinding{
```
mprahl, line 119: You should be able to add the role in `ClusterRoles` when defining the `OCPUser` and the `ClusterRoleBinding` should be created for you.

gparvin: Isn't that only for an existing `ClusterRole`, it needs a new one to have permissions for the managed cluster set binding. @mprahl

gparvin: ah, Thanks!  Will try that soon!

gparvin: Made the update.  Thanks

test/policy-collection/policy_generator_acm_hardening.go
```diff
@@ -0,0 +1,256 @@
...
+	const clustersetRoleName = "grc-e2e-clusterset-role"
+	const subAdminBinding = "open-cluster-management:subscription-admin"
+	ocpUser := common.OCPUser{
+		ClusterRoles: []types.NamespacedName{
+			{Name: "open-cluster-management:admin:local-cluster"},
+			{
+				Name:      "admin",
+				Namespace: namespace,
+			},
+			{Name: "clustersetRoleName"},
```
mprahl, line 64: Did you mean this?
```suggestion
			{Name: clustersetRoleName},
```

gparvin: I wonder if I need to test on a clean system :-)

mprahl (approved):
/hold for other reviews

PR #299, opened 2022-03-21
Add policy-cis-scan test case

ycao56:
Extract compliance operator scan test code and add cis scan test case.
closes https://github.com/stolostron/backlog/issues/20849


gparvin (approved):
Thanks!

PR #302, opened 2022-03-21
Skip publish job checks when image isn't found

dhaiducek:


build/periodic.sh
```diff
@@ -122,63 +133,65 @@ for repo in $REPOS; do
...
+		# check the prow job history
+		checkProwJob "$repo" "$release" "$HAS_IMAGE"
+		if [ $? -eq 1 ]; then
+			rc=1
+		fi
 	done
 done
 
-cleanup
+# cleanup
```
ycao56, line 194: Is this intended?

ycao56:
Are the submodules intended?

build/periodic.sh
dhaiducek: Nope! 😬

dhaiducek:
Shoot--definitely not. Thanks for catching these! I should have looked at it more closely!

ycao56 (approved):

PR #304, opened 2022-03-22
Restructure tests

dhaiducek:
- Use variable for policy-collection base URLs
- Move some common functions to integration_suite_test.go

### DIFF
--- a/test/integration/policy_comp_operator_test.go
+++ b/test/integration/policy_comp_operator_test.go
@@ -34,39 +34,6 @@ func isOCP46andAbove() bool {
 	// should be ocp 4.6 and above
 	return true
 }
-
-var (
-	canCreateOpenshiftNamespacesInitialized bool
-	canCreateOpenshiftNamespacesResult      bool
-)
-
-func canCreateOpenshiftNamespaces() bool {
-	// Only check once - this makes it faster and prevents the answer changing mid-suite.
-	if canCreateOpenshiftNamespacesInitialized {
-		return canCreateOpenshiftNamespacesResult
-	}
-
-	canCreateOpenshiftNamespacesInitialized = true
-	canCreateOpenshiftNamespacesResult = false
-
-	// A server-side dry run will check the admission webhooks, but it needs to use the right
-	// serviceaccount because the kubeconfig might have superuser privileges to get around them.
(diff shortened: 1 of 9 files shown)


test/integration/policy_comp_operator_test.go
```diff
@@ -175,8 +175,7 @@ func complianceScanTest(scanPolicyName string, scanPolicyUrl string, scanName st
 }
 
 var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] Test compliance operator and scan", Label("policy-collection", "stable", "BVT"), func() {
-	const policyCollectStableURL = "https://raw.githubusercontent.com/stolostron/policy-collection/main/stable"
-	const compPolicyURL = policyCollectStableURL + "/CA-Security-Assessment-and-Authorization/policy-compliance-operator-install.yaml"
+	const compPolicyURL = policyCollectCAURL + "policy-compliance-operator-install.yaml"
```
ycao56, line 178: Should this file also be moved to folder `CA-Security-Assessment-and-Authorization`?

dhaiducek: I wasn't 100% sure since it's a larger test that includes some CM policies, but it would indeed make sense to move it over--thanks for pointing it out! I'll create some dummy files in the CM folder to direct people over to this file if they're looking for the E8/CIS tests.

test/integration/policy_compliance_operator_cis_scan_test.md
```diff
@@ -0,0 +1 @@
+The test for `policy-compliance-operator-cis-scan.yaml` can be found in the [Compliance Operator](./policy_comp_operator_test.go) test.
```
ycao56, line 1: Are these README files still needed? Not a big deal to me. Just checking.

dhaiducek: I guess I'm on the fence, too. I do think they're helpful for transparency/traceability if someone is wondering whether the policies are tested since they don't fit the patterns of the other policies that have a dedicated test file, but they are extra files in a large list...

ycao56: I think we could either remove these files or just provide one single README to explain what each test case does.

ycao56 (approved):

PR #309, opened 2022-03-24
Add test for `policy-scc.yaml`

dhaiducek:
Notes:
- Reduces package-level variables to prevent collisions and allow for similarly named variables in each test file
- Adds test for `policy-scc.yaml`
  - Along with the package-level variable reduction, with this test file I attempted to establish a boilerplate test format where a new test for a new kind could be written with minimal updates (primarily to the variable declarations at the top) and then built upon

Addresses:
- https://github.com/stolostron/backlog/issues/20846

### DIFF
--- a/test/integration/policy_scc_test.go
+++ b/test/integration/policy_scc_test.go
@@ -1,0 +1,118 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	policiesv1 "github.com/stolostron/governance-policy-propagator/api/v1"
+	"github.com/stolostron/governance-policy-propagator/test/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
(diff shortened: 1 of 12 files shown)


ycao56 (approved):
This is great!

test/integration/policy_scc_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+		By("Patching placement rule")
+		err = common.PatchPlacementRule(
+			userNamespace, "placement-"+rootPolicyName, clusterNamespace, kubeconfigHub,
+		)
+		Expect(err).To(BeNil())
+
+		By("Checking that " + rootPolicyName + " exists on the Hub cluster")
+		rootPolicy := utils.GetWithTimeout(
+			clientHubDynamic, common.GvrPolicy, rootPolicyName, userNamespace, true, defaultTimeoutSeconds,
```
ycao56, line 64: policy name is `policy-securitycontextconstraints`

ycao56: file name is `policy-scc`

dhaiducek: Oh, man! Fixing now. Looks like SCC isn't namespaced, either. And this particular one might be preinstalled on clusters, so it's probably always compliant...

ycao56 (approved):

dhaiducek:
@ycao56 I just noticed that the label on `policy-collection-test` was updated so it would work as intended. Should I add it back or leave it removed?

ycao56:
> @ycao56 I just noticed that the label on the `policy-collection-test` Make target was updated so it would work as intended. Should I add it back or leave it removed?

It's fine to be removed.

PR #312, opened 2022-03-27
Run etcd-encryption at last

ycao56:
If no `GINKGO_LABEL_FILTER` is set, automatically separate the test cases into two groups.

Side effect: It will generate a lot of skipped test cases due to the change and will affect the reporting in canary.

We don't need this change if canary test cases are not affected by etcd encryption.


ycao56:
Closing this PR for now as it seems etcd encryption does affect tests in Canary and Prow. Will reconsider if needed in the future.

PR #319, opened 2022-04-06
PlacementRule is making calls to PlacementDecisions

gparvin:
There are some rbac failures in the placementrules controller that
shows placementdecisions verbs get, list, watch are needed for the
user access of placement.  This may be needed in the tests that
use placementrule too.

Refs:
 - https://github.com/stolostron/backlog/issues/21455


mprahl (approved):

gparvin:
/retest
Looks like the test environment hit some kind of timeout

PR #321, opened 2022-04-08
Downstream builds are running with no vulnerable images

gparvin:
Our container security operator test assumes we'll have a vulnerable
image somewhere.  That's not the case for current downstream 2.5
builds.  Adding a vulnerable image as part of the test case.

Refs:
 - https://github.com/stolostron/backlog/issues/21459


mprahl (approved):

gparvin:
The amended commit includes the `namespace` in the deployment which I forgot to set.

mprahl (approved):

PR #338, opened 2022-04-27
Update the slack token used to publish test results from travis

gparvin:
The previous token from Yu is no longer working of course.  Need to
update it to a new one I created.

Note that the command to encode the token is what reformatted the file. 


ChunxiAlexLuo (approved):
/lgtm

PR #350, opened 2022-05-05
Use the KUBECONFIG env var when running oc login

mprahl:
In the 2.5 BVT, an issue often appears where the Policy Generator test
runs and then subsequent tests fail to update PlacementRule objects
because the PlacementRule controllers says that it was done by the now
deleted Policy Generator test temporary user.

My theory is that `--kubeconfig` on the `oc login` isn't enough
sometimes, so hopefully adding the `KUBECONFIG` environment variable
fixes the issue.

Related:
https://github.com/stolostron/backlog/issues/22232

Signed-off-by: mprahl <mprahl@users.noreply.github.com>


JustinKuli (approved):
It seems odd that this could work, but it's worth a try.

PR #351, opened 2022-05-06
Test results when a failure happens in an AfterAll

JustinKuli:
This is just a PR to run the CI in prow to see what gets reported in the integration test XML report.


JustinKuli:
/test test-e2e-grc-framework

The previous failure checks what happens when something fails in an `AfterAll` block: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/stolostron_governance-policy-framework/351/pull-ci-stolostron-governance-policy-framework-main-test-e2e-grc-framework/1524432384265555968/artifacts/test-e2e-grc-framework/e2e/artifacts/integration.xml

The failure is logged as a failed testcase, but not as a separate testcase - it is reported on the status of the last (otherwise successful) testcase. In this example, the `The prod Namespace should exist` testcase. And since fail-fast mode is on, all the other tests are skipped.

Now we'll see how it's reported in a `BeforeAll` block.

JustinKuli:
This failure demonstrates what happens when an assertion in a BeforeAll fails: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/stolostron_governance-policy-framework/351/pull-ci-stolostron-governance-policy-framework-main-test-e2e-grc-framework/1524465595054034944/artifacts/test-e2e-grc-framework/e2e/artifacts/integration.xml

Similar to the AfterAll, it doesn't create a separate testcase, the failure gets attached to another testcase. In this case, the test that was _about_ to run (the first It in the Describe block).

dhaiducek:
Cool--thanks for confirming, @JustinKuli !

PR #362, opened 2022-05-18
BVT failing in same testcase but different test times out

gparvin:
The same generator test is timing out waiting for the policyset to be
created.  Previously the timeout was waiting for the policyset to
propagate to the managed cluster, which is a later test.  Suspect
test environments are just causing this.

Refs:
 - https://github.com/stolostron/backlog/issues/22546


gparvin:
The cluster claim isn't working.  Will try a retest again
/retest

gparvin:
The cluster claims aren't working right now for this pr so ignoring the test.  The change is simple so I think this is fine.

JustinKuli (approved):

PR #367, opened 2022-05-24
Set release branch to a variable

dhaiducek:
We'll need to update the `api` commit hash each time we update packages, but this will at least make it so that we won't have to update the release branch each time.


dhaiducek:
For reference, `PULL_BASE_REF` is a variable available in Prow:
- https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
- https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

gparvin (approved):

PR #370, opened 2022-05-25
Hibernate E2E cluster when not in use

dhaiducek:
This implements a hibernation flow that wakes up the `grce2e-policy-grc-cp-prow` ClusterClaim prior to the tests and hibernates it on completion.

Related to:
- https://github.com/stolostron/backlog/issues/14845

### DIFF
--- a/build/set-claim.sh
+++ b/build/set-claim.sh
@@ -1,0 +1,44 @@
+#! /bin/bash
+
+# Log into Collective cluster
+KUBECONFIG_FILE="${PWD}/kubeconfig-collective"
+touch ${KUBECONFIG_FILE}
+export KUBECONFIG=${KUBECONFIG_FILE}
+
+oc login --token=${COLLECTIVE_TOKEN} https://api.collective.aws.red-chesterfield.com:6443 &>/dev/null
+if [ $? = "0" ]; then
+  echo "Logged in to Collective cluster"
+else
+  echo "Failed to log in to Collective cluster"
+  exit 1
+fi
+
+# Set E2E ClusterClaim PowerState (Running/Hibernating)
+CLAIM="grce2e-policy-grc-cp-prow"
+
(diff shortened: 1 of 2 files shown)


build/set-claim.sh
```diff
@@ -0,0 +1,44 @@
+#! /bin/bash
+
+# Log into Collective cluster
+KUBECONFIG_FILE="${PWD}/kubeconfig-collective"
+touch ${KUBECONFIG_FILE}
+export KUBECONFIG=${KUBECONFIG_FILE}
+
+oc login --token=${COLLECTIVE_TOKEN} https://api.collective.aws.red-chesterfield.com:6443 &>/dev/null
```
JustinKuli, line 8: Does this token never expire?

build/set-claim.sh
```diff
@@ -0,0 +1,44 @@
...
+if [ $? = "0" ]; then
+  echo "Logged in to Collective cluster"
+else
+  echo "Failed to log in to Collective cluster"
+  exit 1
+fi
+
+# Set E2E ClusterClaim PowerState (Running/Hibernating)
+CLAIM="grce2e-policy-grc-cp-prow"
```
JustinKuli, line 17: This name will change if we re-create the cluster, is that right? I think this is fine for now, but maybe if we are fully recreating that cluster more often in the future, we could consider using a label or something.

dhaiducek: I thought of that. I figure eventually we might use this script to provision claims. But since it's a claim, we should be able to reuse the name since we'd be deleting and recreating it. I'm not sure why the claim name ever changed in the past--that was always odd to me.

JustinKuli: When I recreated the cluster, I gave the claim a new name, that way I could spin up the new cluster while the other one still existed. `¯\_(ツ)_/¯`

JustinKuli (approved):

PR #371, opened 2022-05-25
Patch images in Kind

dhaiducek:
We pull manifests from each repo, which specifies the `latest` tag. This adds patching logic so that we make sure we're pulling the `latest-<version>` tag.


gparvin (approved):

dhaiducek:
That cherry pick wouldn't have worked, anyway with the PolicySet failures. Oops. 🤦

PR #373, opened 2022-05-26
Test improvements

JustinKuli:
DRAFT - testing (hopefully) the Travis parallelization settings.


JustinKuli:
Previous kind Travis tests (running sequentially) completed in about 18 minutes... The new ones, in parallel, completed in about 14 minutes! A lot of the time in the "test" is doing the setup of kind and OLM.

Locally, times went from about 13 minutes in the ginkgo tests to 6 minutes (this does not count kind setup).

The prow tests fail because I need to figure out how to install `oc` there I think.

PR #377, opened 2022-05-27
Use script instead of Make target

dhaiducek:
It's not working, but I shouldn't have relied on the build-harness when we have a script that installs `oc` already.


JustinKuli (approved):
And that script should work in Travis! But from my PR, I know it doesn't quite work in Prow, FYI.

PR #381, opened 2022-06-01
Use Prow variable for dynamic image tag

dhaiducek:
This addresses the "cut a new release" script deficiency by allowing us to remove the environment variable from Prow and set it dynamically using a Prow variable, similar to this recently implemented code: https://github.com/stolostron/governance-policy-framework/blob/8006a5e88d3649647f3202caf6c1a739c2786d52/build/run-test-kind-prow.sh#L36-L39

### DIFF
--- a/build/run-e2e-tests-policy-framework-prow.sh
+++ b/build/run-e2e-tests-policy-framework-prow.sh
@@ -27,6 +27,13 @@ echo "* Install cert manager"
 $DIR/install-cert-manager.sh
 
 echo "* Set up cluster for test"
+# Set tag for images: Use `latest` for `main` and `latest-<version>` for `release-<version>` branches
+# `PULL_BASE_REF` is a variable provided by Prow: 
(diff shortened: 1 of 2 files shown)


build/run-e2e-tests-policy-framework-prow.sh
```diff
@@ -27,6 +27,10 @@ echo "* Install cert manager"
 $DIR/install-cert-manager.sh
 
 echo "* Set up cluster for test"
+export VERSION_TAG="latest"
+if [ "${PULL_BASE_REF}" ] && [ "${PULL_BASE_REF}" != "main" ]; then
+  export VERSION_TAG="${VERSION_TAG}-${PULL_BASE_REF#*-}"
```
JustinKuli, line 32: Can you add a comment (or something) with an example of what the result will be from this? I think it's `latest-2.4`, but I'm not actually positive.

JustinKuli (changes requested):

JustinKuli (approved):

PR #382, opened 2022-06-02
Update image tag for `openshift/release` PRs

dhaiducek:


build/run-e2e-tests-policy-framework-prow.sh
```diff
@@ -27,13 +27,20 @@ echo "* Install cert manager"
...
+# Otherwise, use `PULL_BASE_REF`.
+VERSION_TAG="latest"
+if [[ "${REPO_OWNER}" == "openshift" ]] && [[ "${REPO_NAME}" == "release" ]]; then
+  TARGET_BRANCH="$(echo "${JOB_SPEC}" | jq -r '.extra_refs[] | select(.workdir == true).base_ref')"
+else
+  TARGET_BRANCH="${PULL_BASE_REF}"
 fi
+if [ "${TARGET_BRANCH}" ] && [ "${TARGET_BRANCH}" != "main" ]; then
+  VERSION_TAG="${VERSION_TAG}-${PULL_BASE_REF#*-}"
+fi
```
JustinKuli, line 42: Nit: I know it was single `[` before, but since the new part uses `[[`, I'd rather use that consistently. Unless there's a good reason?

I think you've made and maintained more of these scripts than I have, so if you have another style guide then lmk, but I usually go to https://google.github.io/styleguide/shellguide.html

JustinKuli (dismissed):
/hold for optional nit

build/run-e2e-tests-policy-framework-prow.sh
dhaiducek: Nope, no style guide--that looks like a good reference, thanks! I definitely prefer consistency, and `[[` certainly is much nicer to deal with. The one minor thing is that it's my understanding that `[` is POSIX compliant while `[[` is not, but in this case it probably doesn't matter. I'll update it. `[[` fo life! 😆

JustinKuli (approved):

PR #386, opened 2022-06-07
Add qualified policy type name and pin gatekeeper operator version

JustinKuli:
Kyverno policies are also just called "policy", so the tests need to be
a little more specific when dealing with open-cluster-management.io
policies through kubectl commands.

In addition, the gatekeeper operator `latest` image was recently updated, and we need to pin to a specific version in order to keep it working.

PR #387, opened 2022-06-08
Creation of deployment from subscription can take some time

gparvin:
The deployment gets created when the subscription installs the kyverno
helm chart.  Sometimes this can take a while for the helm install to
start so the default timeout is really a bit short.  Increasing it to
reduce the concern here.


JustinKuli:
@gparvin, I'm going to bring this into my https://github.com/stolostron/governance-policy-framework/pull/386 - I don't think your PR can pass without the gatekeeper fix in mine, and mine probably can't pass without the change here 😂

PR #389, opened 2022-06-09
Gk audit timing

JustinKuli:
The audit pod is taking a long time to complete and populate a field that the sample gatekeeper policy is checking for. This increases the timeout on that test, and should gather some information we can use to track how long the process is taking as the cluster ages.

### DIFF
--- a/test/integration/policy_gatekeeper_operator_downstream_test.go
+++ b/test/integration/policy_gatekeeper_operator_downstream_test.go
@@ -202,6 +202,17 @@ var _ = Describe("RHACM4K-3055", Ordered, Label("policy-collection", "stable", "
 				return details[1].(map[string]interface{})["compliant"]
 			}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
 		})
+		It("Grabs the gatekeeper audit duration metric", func() {
+			auditPodName, _ := common.OcManaged("get", "pod", "-n=openshift-gatekeeper-system",
+				"-l=control-plane=audit-controller", `-o=jsonpath={.items[0].metadata.name}`)
+			common.OcManaged("exec", "-n=openshift-gatekeeper-system", auditPodName, "--", "bash",
+				"-c", "curl -s localhost:8888/metrics | grep -A1 audit_duration_seconds_sum")
+			/* example output:
+			gatekeeper-audit-9c88bf969-mgf5r
+			gatekeeper_audit_duration_seconds_sum 1005.4185594219999
+			gatekeeper_audit_duration_seconds_count 25
+			*/
+		})
 		It("Creating a valid ns should not be blocked by gatekeeper", func() {
 			By("Creating a namespace called e2etestsuccess on managed")
 			Eventually(func() interface{} {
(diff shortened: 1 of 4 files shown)


test/integration/policy_gatekeeper_operator_downstream_test.go
```diff
@@ -193,14 +194,19 @@ var _ = Describe("RHACM4K-3055", Ordered, Label("policy-collection", "stable", "
 		})
 		It("stable/policy-gatekeeper-sample should be compliant", func() {
 			By("Checking if the status of root policy is compliant")
-			Eventually(getComplianceState(GKPolicyName), defaultTimeoutSeconds*12, 1).Should(Equal(policiesv1.Compliant))
+			Eventually(getComplianceState(GKPolicyName), 10*time.Minute, 1).Should(Equal(policiesv1.Compliant))
```
dhaiducek, line 197: Now, I know what this means! 😄

test/integration/policy_gatekeeper_operator_downstream_test.go
```diff
@@ -193,14 +194,19 @@ var _ = Describe("RHACM4K-3055", Ordered, Label("policy-collection", "stable", "
...
 				plc := utils.GetWithTimeout(clientHubDynamic, common.GvrPolicy, userNamespace+"."+GKPolicyName, clusterNamespace, true, defaultTimeoutSeconds)
 				details := plc.Object["status"].(map[string]interface{})["details"].([]interface{})
 				return details[1].(map[string]interface{})["compliant"]
 			}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
 		})
+		It("Grabs gatekeeper audit logs to learn about gatekeeper audit duration", func() {
+			_, err := common.OcManaged("logs", "-n=openshift-gatekeeper-system", "-l=control-plane=audit-controller", "--tail=-1")
+			Expect(err).To(BeNil())
+			// grep for 'event_type":"audit' to get the interesting bits.
+		})
```
dhaiducek, line 209: Nice! I've realized that our debug logs are lacking because we clean up resources after each test, and it might be nice to have a common debug function that grabs policy objects to be deleted and logs similar to this for pods that aren't GRC. 
Should this output to a file, though? I'm not sure but there might be limits on the log length in Travis (and elsewhere), but it'll be included in the log dump if it's added to the debug directory.

JustinKuli: I agree, currently the cleanup and debug logs aren't playing nicely. A common debug function could be a useful solution.

I wasn't too worried about the log length, it's really not much compared to what ginkgo is outputting anyway... but I was looking at the metrics anyway so I put in another way to get the information I was looking for. The output should be smaller, although the commands are a bit more "hacky".

dhaiducek (approved):

test/integration/policy_gatekeeper_operator_downstream_test.go
dhaiducek: Hacky indeed, but I like the shorter more targeted approach. We can leave lengthier logs for a common debug function.

gparvin (approved):
THanks for debugging this more.

PR #390, opened 2022-06-13
Disable cluster hibernation

dhaiducek:


JustinKuli:

Let's see if this works... this change wouldn't mess with the tests in prow

dhaiducek:
Thanks! (The Prow failure is due to some mismatched status checking due to the Collective upgrade.)

JustinKuli:
Ah cool, I only took a 5 second look, and figured the clusterpool was empty. So we need to update logic for the clusterpool stuff?

PR #393, opened 2022-06-22
Manually add deprecated UI repos to CI check

dhaiducek:
There was a minor update to `grc-ui` that failed in CI but got missed since it was removed from the repo list. This manually adds deprecated repos to our script.


ChunxiAlexLuo (commented):
Is this a temporary solution? 

So this change targets 2.6 and future releases but `grc-ui-api` is still owned by the console squad？ Just confirm

dhaiducek:
No. This change targets all releases and isn't tied to a release branch in particular. We now own all of our UI maintenance and updates (current and past releases). 
The script will skip the check for a particular repo if the release branch isn't found. The script already checks `2.3` through `2.5`, but that's a good point--we need to add `2.6`! I'll add a commit to add it.

ChunxiAlexLuo (approved):
/lgtm

PR #394, opened 2022-06-24
Increase policy_comp_operator_test timeout

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo <chuluo@redhat.com>

https://github.com/stolostron/backlog/issues/23684

These test cases failed due to operator status delay.  PENDING vs RUNNING / PENDING vs AGGREGATING, etc

Increase policy_comp_operator_test timeout and continuously observe it.


gparvin (approved):
With a really long timeout I'm not sure if we could cause some canary issues, but hopefully these issues will be rare.  Thanks for looking into this more!

PR #398, opened 2022-07-11
Improve policy_governance_info metric RBAC tests

JustinKuli:
In newer versions of Kubernetes, tokens are not automatically created
for all service accounts. Therefore, the tests might need to manually
create the tokens for the service accounts it uses.

There is also a new test for users that should not have access to the
endpoint, whereas previously we only tested unauthenticated access.

Refs:
 - https://github.com/stolostron/backlog/issues/24075

### DIFF
--- a/test/integration/policy_info_metric_test.go
+++ b/test/integration/policy_info_metric_test.go
@@ -5,7 +5,6 @@ package integration
 import (
 	"context"
 	"encoding/base64"
-	"fmt"
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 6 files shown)


JustinKuli:
/unhold
Ok, now it should work on OCP 4.11 - I didn't realize that tokens don't automatically get added to the account's `secrets` list. We don't need it there anyway, we can just always create and use our own token.

test/integration/policy_info_metric_test.go
```diff
@@ -71,37 +78,63 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test policy_governance_info metric
...
+		_, err = common.OcHub("apply", "-f", noMetricsTokenYaml, "-n", userNamespace)
 		Expect(err).To(BeNil())
-		_, err = common.OcHub("create", "clusterrolebinding", roleBindingName, "--clusterrole=cluster-admin", fmt.Sprintf("--serviceaccount=%s:%s", userNamespace, saName))
+
+		var encodedtoken string
+
+		// The secret could take a moment to be populated with the token
+		Eventually(func(g Gomega) {
+			var err error
+			encodedtoken, err = common.OcHub("get", "secret", noMetricsTokenName,
```
mprahl, line 94: Optional: It'd be cleaner to use the Go client and then you wouldn't have to check the length of the string to determine if the token is present or not.

mprahl (approved):
/hold for optional comment and/or other review

willkutler (approved):

PR #401, opened 2022-07-14
Move the policy-imagemanifestvuln test to SVT

mprahl:
This test is flaky and the errors are almost never indicative of an
actual issue with the GRC code. This should be an SVT test instead for
those reasons.

### DIFF
--- a/test/integration/policy_imagemanifest_test.go
+++ b/test/integration/policy_imagemanifest_test.go
@@ -17,7 +17,7 @@ import (
 	"github.com/stolostron/governance-policy-framework/test/common"
 )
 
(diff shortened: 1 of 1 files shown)


JustinKuli:
Seems good to me, but can I defer to @gparvin for the final decision? Maybe there is a reason I'm missing for why this was in the BVT to start with

ChunxiAlexLuo (approved):
/lgtm

gparvin (approved):
I agree on taking this out of BVT.  We've been consistently removing from BVT the tests that install operators related to other products.  Originally they were all part of BVT because we defaulted all of our stable policy testing to BVT.  I think we still get good coverage for stable policies in SVT.

PR #403, opened 2022-07-15
Move the policy-imagemanifestvuln test to SVT

mprahl:
This test is flaky and the errors are almost never indicative of an
actual issue with the GRC code. This should be an SVT test instead for
those reasons.

This also cherry-picks some changes to make the CI pass.

This is backporting #401.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,7 @@
 TRAVIS_BUILD ?= 1
 
 PWD := $(shell pwd)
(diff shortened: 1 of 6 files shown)


gparvin (approved):

gparvin (approved):

mprahl:
@gparvin could you please review this again?

PR #408, opened 2022-07-22
Add tests for pruneObjectBehavior feature

JustinKuli:
Tests the general functionality of this new feature.

These tests are in a new (experimental) format that allow them to
be run in our usual KinD tests (under e2e) as well as in our larger set
of tests that run against real clusters (under integration). This
format would also potentially allow us to turn tests on and off
separately from the ginkgo filters, which could be helpful because the
filters cause tests to be *skipped*, which impacts test execution stats.

Refs:
 - https://github.com/stolostron/backlog/issues/24440

### DIFF
--- a/test/common/gvr.go
+++ b/test/common/gvr.go
@@ -7,6 +7,7 @@ import "k8s.io/apimachinery/pkg/runtime/schema"
 var (
 	GvrPod                   = schema.GroupVersionResource{Version: "v1", Resource: "pods"}
 	GvrNS                    = schema.GroupVersionResource{Version: "v1", Resource: "namespaces"}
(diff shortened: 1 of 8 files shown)


JustinKuli:
The tests take about 4 minutes to run, so I don't think they are a great fit in the BVT. It also feels a little lengthy for the KinD tests, but it is a core functionality of the config-policy-controller...

JustinKuli:
ocm-e2e-clusterpool-cluster-deploy log says: `Using snapshot 2.6.0-SNAPSHOT-2022-07-13-23-23-00` so that's just not new enough

JustinKuli:
Pulled the test out of the "integration" tests, because we need a working `dev` release to have the backend changes.

willkutler:
/lgtm

great work @JustinKuli !!

PR #411, opened 2022-07-27
Implement test for foreground pruning

JustinKuli:
The new test does not check for the Policy on the managed cluster, but
in the future we may implement something to keep that around (possibly
also on the hub) while the ConfigurationPolicy is cleaning up.

Refs:
 - https://github.com/stolostron/backlog/issues/24440


gparvin (approved):
This is great!  Thanks for the nice foreground deletion test.

PR #415, opened 2022-07-29
Zts cmc 2

mahesh-zetta:


mahesh-zetta:
https://github.com/stolostron/governance-policy-framework/pull/416

Replaced it with above PR that had less commits to be signed off on

PR #416, opened 2022-07-29
Add Test Case For Zts-Cmc Policy

mahesh-zetta:
Signed-off-by: Mahesh Bhoothapuri <mahesh@zettaset.com>

### DIFF
--- a/test/integration/policy_zts_cmc_test.go
+++ b/test/integration/policy_zts_cmc_test.go
@@ -1,0 +1,120 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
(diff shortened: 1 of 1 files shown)


gparvin (approved):
Thanks for working with us to create the policy and the test!

gparvin:
It was a policyreport test that failed the validation -- I think it might be the same one that failed in one of my other PRs so I suspect it's something in our framework -- not this new test.
/retest

gparvin:
Failed to get a cluster... I'll retest later

mahesh-zetta:
> Failed to get a cluster... I'll retest later

The error does not seem to be related to the new test I added.   Please let me know if I need to provide any info.

gparvin:
I think it's other tests
/retest

gparvin:
I think our CI is broken for this test until https://github.com/stolostron/governance-policy-framework/pull/418 merges.  Will need to try this again afterwards.

gparvin:
The test keeps failing in our framework.  It just looks like we aren't able to reserve a cluster for testing.
/retest

gparvin:
@dhaiducek I keep getting failures to get a cluster to test on.  Any idea why this PR would keep getting those failures?  Do I just need to keep trying or is something broken somewhere?

dhaiducek:
This is expected. The `policy-grc-cp-prow` ClusterPool has been exhausted, and the timeout in the Prow workflow isn't long enough for a new cluster to be created. You'll need to check the ClusterPool to be sure there are clusters before restarting the test. (You can also bump the ClusterPool size if you feel you might need to retest again soon.)

PR #417, opened 2022-08-01
Add debug logs when the Kyverno test fails

mprahl:
This should help debug issues such as:
https://github.com/stolostron/backlog/issues/24699

Closes https://github.com/stolostron/backlog/issues/24699

### DIFF
--- a/test/common/common.go
+++ b/test/common/common.go
@@ -258,3 +258,19 @@ func DoRootComplianceTest(hub dynamic.Interface, policyName string, compliance p
 		1,
 	).Should(gomega.Equal(compliance))
 }
+
+func OutputDebugInfo(testName string, additionalResources ...string) {
+	ginkgo.GinkgoWriter.Printf("%s test Kubernetes info:\n", testName)
+
+	resources := []string{
+		"policies.policy.open-cluster-management.io",
+		"placementrules.apps.open-cluster-management.io",
+		"placements.cluster.open-cluster-management.io",
+		"placementbindings.policy.open-cluster-management.io",
+	}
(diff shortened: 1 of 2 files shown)


test/common/common.go
```diff
@@ -258,3 +258,11 @@ func DoRootComplianceTest(hub dynamic.Interface, policyName string, compliance p
 		1,
 	).Should(gomega.Equal(compliance))
 }
+
+func OutputDebugInfo(testName string) {
+	ginkgo.GinkgoWriter.Printf("%s test Kubernetes info:\n", testName)
+
+	for _, resource := range []string{"policies", "placementrule", "placement", "placementbinding"} {
```
JustinKuli, line 265: I think we should qualify the "policies" name - do you want kyverno policies, open-cluster-management policies, maybe both?

JustinKuli (changes requested):

test/common/common.go
```diff
@@ -258,3 +258,19 @@ func DoRootComplianceTest(hub dynamic.Interface, policyName string, compliance p
 		1,
 	).Should(gomega.Equal(compliance))
 }
+
+func OutputDebugInfo(testName string, additionalResources []string) {
```
JustinKuli, line 262: Nit: could be variadic for the additionalResources - currently to not get any additional resources, you need to pass `nil`. But this is totally not necessary right now.

JustinKuli (dismissed):

test/common/common.go
mprahl: That's a nice suggestion!

JustinKuli (approved):

PR #331, opened 2022-04-20
Use Ordered Describes and After/BeforeAll nodes

JustinKuli:
Since fail-fast was turned on for the tests, some of the "clean up" test
nodes are not being run. A better fit for those is "AfterAll" nodes,
which require the Describes to be marked as Ordered.

In addition, some BeforeEach test nodes would fit better as BeforeAll,
since that is now possible.

Note: adding "Ordered" to the Describes largely does not change the
behavior. Unless the `--randomize-all` flag is passed, the It nodes
will be run in the order they are defined anyway.

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

PR #419, opened 2022-08-03
Use Ordered Describes and After/BeforeAll nodes (#331)

gparvin:
Backport of PR https://github.com/stolostron/governance-policy-framework/pull/331 to release-2.5 for issue:
- https://github.com/stolostron/backlog/issues/23871

Since fail-fast was turned on for the tests, some of the "clean up" test
nodes are not being run. A better fit for those is "AfterAll" nodes,
which require the Describes to be marked as Ordered.

In addition, some BeforeEach test nodes would fit better as BeforeAll,
since that is now possible.

Note: adding "Ordered" to the Describes largely does not change the
behavior. Unless the `--randomize-all` flag is passed, the It nodes
will be run in the order they are defined anyway.

### DIFF
--- a/test/e2e/hub_templates_encryption_test.go
+++ b/test/e2e/hub_templates_encryption_test.go
@@ -20,7 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-var _ = Describe("Test Hub Template Encryption", func() {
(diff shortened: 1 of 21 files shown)


gparvin:
Latest test failure was no clusters were obtained to run the test.  Will kick this off again tomorrow.

JustinKuli (approved):
Feels strange to approve my own cherry-picked commit 😂

PR #425, opened 2022-08-10
Fix assertions wrapped in Eventuallys

mprahl:
When an assertion is made in an `Eventually`, it must use the input
`Gomega` argument to avoid a failed assertion from failing the
`Eventually` before it times out.

One example is a flaky network connection failure an SVT test:
https://github.com/stolostron/backlog/issues/24895

### DIFF
--- a/test/e2e/gatekeeper_test.go
+++ b/test/e2e/gatekeeper_test.go
@@ -144,10 +144,10 @@ var _ = Describe("Test gatekeeper", Ordered, func() {
 				details := plc.Object["status"].(map[string]interface{})["details"].([]interface{})
 				return details[1].(map[string]interface{})["history"]
 			}, defaultTimeoutSeconds, 1).ShouldNot(BeNil())
-			Eventually(func() interface{} {
+			Eventually(func(g Gomega) interface{} {
 				plc := utils.GetWithTimeout(clientHubDynamic, common.GvrPolicy, userNamespace+"."+GKPolicyName, clusterNamespace, true, defaultTimeoutSeconds)
 				details := plc.Object["status"].(map[string]interface{})["details"].([]interface{})
-				Expect(details[1].(map[string]interface{})["history"]).NotTo(BeNil())
+				g.Expect(details[1].(map[string]interface{})["history"]).NotTo(BeNil())
 				return details[1].(map[string]interface{})["history"].([]interface{})[0].(map[string]interface{})["message"]
 			}, defaultTimeoutSeconds, 1).Should(Equal("Compliant; notification - k8srequiredlabels [ns-must-have-gk] found as specified, therefore this Object template is compliant"))
 			By("Checking if violation message for policy template policy-gatekeeper-admission is compliant")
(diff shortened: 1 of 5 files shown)


JustinKuli:
It feels like this should work, but this line in the docs worries me: https://pkg.go.dev/github.com/onsi/gomega@v1.20.0#Eventually

> Eventually also supports a special case polling function that takes a single Gomega argument and returns no values

mprahl:
> It feels like this should work, but this line in the docs worries me: https://pkg.go.dev/github.com/onsi/gomega@v1.20.0#Eventually
> 
> > Eventually also supports a special case polling function that takes a single Gomega argument and returns no values

I was modeling this after "Category 3" in the docs, though it is peculiar how it also returns an error in that example.

dhaiducek:
Apparently errors are failures. I'm wondering if this could be leveraged to simplify our tests somewhere?
>If multiple values are returned by the function, Eventually will pass the first value to the matcher and require that all others are zero-valued. This allows you to pass Eventually a function that returns a value and an error - a common pattern in Go.
>
>For example, consider a method that returns a value and an error:
>```
>func FetchFromDB() (string, error)
>```
>Then
>```
>Eventually(FetchFromDB).Should(Equal("got it"))
>```
>will pass only if and when the returned error is nil *and* the returned string satisfies the matcher.

dhaiducek:
Ah, indeed we can! It looks like the internal `Expect` on the error isn't actually necessary.

mprahl:
> Ah, indeed we can! It looks like the internal `Expect` on the error isn't actually necessary.

In this PR, I don't see a case where the `Expect` on the error isn't followed by returning a different value then the function call. So I think `Expect` on the error is better since the alternative would be something like this:
```go
Eventually(func(g Gomega) interface{} {
	podList, err := clientManaged.CoreV1().Pods("openshift-gatekeeper-operator").List(context.TODO(), metav1.ListOptions{LabelSelector: "control-plane in (controller-manager, gatekeeper-operator-controller-manager)"})
	g.Expect(err).To(BeNil())
	return len(podList.Items)
}, defaultTimeoutSeconds*8, 1).ShouldNot(Equal(0))
```

To:

```go
Eventually(func(g Gomega) (interface{}, error) {
	podList, err := clientManaged.CoreV1().Pods("openshift-gatekeeper-operator").List(context.TODO(), metav1.ListOptions{LabelSelector: "control-plane in (controller-manager, gatekeeper-operator-controller-manager)"})
	if err != nil {
		return interface{}{}, err
	}
	return len(podList.Items), nil
}, defaultTimeoutSeconds*8, 1).ShouldNot(Equal(0))
```

dhaiducek:
Oh yeah. I think I was hoping something like this would work:
```golang
Eventually(func() (*unstructured.UnstructuredList, error) {
	return clientManaged.CoreV1().Pods("openshift-gatekeeper-operator").List(context.TODO(), metav1.ListOptions{LabelSelector: "control-plane in (controller-manager, gatekeeper-operator-controller-manager)"})
}, common.MaxTravisTimeoutSeconds, 1).ShouldNot(HaveKeyWithValue("Items", HaveLen(0)))
```
But it's not like that's all that different/better. (And I'm not even sure if it'd work!)

mprahl:
@dhaiducek and @JustinKuli I'd like to keep it as is since I don't see an improvement in the alternatives.

Could you please both review it again?

JustinKuli (approved):

PR #427, opened 2022-08-11
Pin the oc client version to 4.10.21

JustinKuli:
Some recent failures have been caused by the openshift client version;
when a test exposes a service as a route, it provides an `--overrides`
paramter that was being ignored in newer client versions.

### DIFF
--- a/build/download-clis.sh
+++ b/build/download-clis.sh
@@ -9,7 +9,7 @@ set -e
 if ! which oc > /dev/null; then
     echo "Installing oc and kubectl clis..."
     mkdir clis-unpacked
-    curl -kLo oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz
+    curl -kLo oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.10.21/openshift-client-linux.tar.gz
     tar -xzf oc.tar.gz -C clis-unpacked
     chmod 755 ./clis-unpacked/oc
     chmod 755 ./clis-unpacked/kubectl
--- a/build/run-e2e-tests.sh
+++ b/build/run-e2e-tests.sh
@@ -14,7 +14,7 @@ fi
 if ! which oc > /dev/null; then
     echo "Installing oc and kubectl clis..."
     mkdir clis-unpacked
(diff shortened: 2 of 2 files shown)


JustinKuli:
Alternatively, we might consider updating the commands that create the route to use `oc create route reencrypt --service=<name>` instead of `oc expose service <name> --overrides=...`. But I haven't fully tested that yet

dhaiducek:
Is `--overrides` being ignored or does the JSON need to be wrapped in single quotes--have you experimented with that?

JustinKuli:
`--overrides` in `oc expose` seems like it has been ignored completely, since Create a route based on service nginx.

dhaiducek:
I'm not so sure. It looks like it's there:
- https://github.com/openshift/oc/commit/f036b597cea4fc29f6feb04c6338a41c357022da#diff-d1f533e120ff29e83d9b8f703d188a89abec948c15e5913a9ed38d8393b76668R69
- https://github.com/kubernetes/kubectl/blob/a13a3bab46b7d41fbfcf3af1452cca8537a70c93/pkg/cmd/expose/expose.go#L91

Have you tried the command locally with the latest?

mprahl:
> 

This was from my tests locally:

```
~ ❯ oc version                                                                                          
Client Version: 4.11.0
Kustomize Version: v4.5.4
Server Version: 4.10.21
Kubernetes Version: v1.23.5+3afdacb
~ ❯ oc expose service insights-metrics -n open-cluster-management --overrides='{"apiVersion": "route.o}'
route.route.openshift.io/insights-metrics exposed
~ ❯ ~/bin/oc-4-9 expose service insights-metrics -n open-cluster-management --overrides='{"apiVersion": "route.o}'
error: Invalid JSON Patch
```

JustinKuli:
It's broken because previously `oc` created a generator for the route, which was able to inherit more options from `kubectl expose`. But now it creates the route directly: https://github.com/openshift/oc/blob/master/pkg/cli/expose/expose.go#L187

JustinKuli:
/hold

Alternate PR: https://github.com/stolostron/governance-policy-framework/pull/428

mprahl (approved):

JustinKuli:
/unhold

Per conversation with @mprahl, pinning the version is good for a variety of reasons, and it will get things fixed for other pending PRs more quickly :D

PR #433, opened 2022-08-16
Pin the kind CLI version

mprahl:
This prevents GitHub rate limiting causing failures in the tests. It
additionally provides a consistent version on each test run to improve the
stability of the tests.

This also makes the run-e2e-tests.sh script fail on HTTP errors.


dhaiducek (dismissed):

mprahl:
@dhaiducek could you please review again? It seems that the OS TravisCI is using has an older version of curl so I had to use a different flag.

dhaiducek (approved):

PR #459, opened 2022-08-30
Use user/IDP patterns for hardening test

dhaiducek:
The assumption in the cleanup is that the secret (and IDP) are named `<username>-user`, but the hardening test didn't follow that pattern, preventing proper cleanup.

Fingers crossed this is the cause of the intermittent subadmin issue since if the hardening tests run first (which doesn't always happen and would explain why it pops up once in a while), they would have created an IDP that wasn't properly cleaned up for the following generator tests.

### DIFF
--- a/test/common/user_management.go
+++ b/test/common/user_management.go
@@ -90,7 +90,7 @@ func GetKubeConfig(server, username, password string) (string, error) {
 
 	output, err := cmd.CombinedOutput()
 	if err != nil {
-		os.Remove(f.Name())
+		os.Remove(kubeconfigPath)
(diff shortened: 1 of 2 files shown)


dhaiducek:
I probably overstated the IDP pattern assumption, found here:
https://github.com/stolostron/governance-policy-framework/blob/f2d79a001d0148baa8158a55fed2632bddbc2d98/test/common/user_management.go#L510

But I'm still hopeful that having a separate/fresh IDP name will help resolve the issue.

dhaiducek:
Nevermind--I was looking in the wrong place. But maybe using a different IDP name would do it?

dhaiducek:
Okay, take 2. Different user/IDP name.

PR #451, opened 2022-08-24
Remove completed compliance scan checks from BVT

dhaiducek:
Our E8 and CIS scan policies do have checks for `AGGREGATING` and `DONE`, so it's probably reckless/irresponsible to remove the checks entirely. However, given the length of time it takes for a scan to complete, this PR removes those checks from the BVT so that the scan checks only run on SVT.

~(I'll need to verify that moving these labels works as expected, but I think it should.)~

PR #469, opened 2022-09-09
Trying a new idea around validating the compliance operator status

gparvin:
We want the scan to be Running and ideally get to aggregating but that seems to time out a lot.  I think if it stays running we can still consider the test OK -- but it could also transition to aggregating or done so this PR will let the final state pass in any of these cases.

Refs:
 - https://github.com/stolostron/backlog/issues/25631

### DIFF
--- a/test/integration/policy_comp_operator_test.go
+++ b/test/integration/policy_comp_operator_test.go
@@ -108,19 +108,19 @@ func complianceScanTest(scanPolicyName string, scanPolicyUrl string, scanName st
 				return len(list.Items)
 			}, common.MaxTravisTimeoutSeconds, 1).ShouldNot(Equal(0))
 		})
-		It("ComplianceSuite "+scanName+" scan results should be AGGREGATING", func() {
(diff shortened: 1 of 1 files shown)


gparvin:
The idea is once we get to the Running state -- do we care if it stays Running and gets to Aggregating and Done?  I'm going with we are happy to know we successfully caused the compliance checks to get to Running -- after that we can wait a while to make sure we don't get some other state -- but even if we do it's not going to be our fault I think.

gparvin:
I didn't do a local run so holding to make sure this looks ok...
/hold

dhaiducek:
@gparvin This never should have run if it was a BVT. I took the BVT label off the completion of the scans. I'll take a peek.

dhaiducek:
Well shoot. I didn't cherry-pick it, sorry:
- https://github.com/stolostron/governance-policy-framework/pull/451

dhaiducek:
It had a followup PR also:
- https://github.com/stolostron/governance-policy-framework/pull/461

gparvin:
Closing in favor of existing PRs 451, and 461

PR #470, opened 2022-09-12
Output the Gatekeeper pods when the deployment fails

mprahl:
This also pins the Kubernetes version to v1.24 for the latest to workaround:
https://github.com/stolostron/backlog/issues/25698

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,13 @@ KIND_MANAGED_NAMESPACE ?= open-cluster-management-agent-addon
 MANAGED_CLUSTER_NAME ?= managed
 HUB_CLUSTER_NAME ?= hub
 KIND_VERSION ?= latest
+
+# TODO: Remove this since this a workaround until the following is addressed.
+# https://github.com/stolostron/backlog/issues/25698
+ifeq ($(KIND_VERSION), latest)
+	KIND_VERSION = v1.24.4
+endif
+
 ifneq ($(KIND_VERSION), latest)
 	KIND_ARGS = --image kindest/node:$(KIND_VERSION)
 else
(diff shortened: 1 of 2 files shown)


test/e2e/gatekeeper_test.go
```diff
@@ -65,6 +65,13 @@ var _ = Describe("Test gatekeeper", Ordered, func() {
 			By("Checking number of pods in gatekeeper-system ns")
 			utils.ListWithTimeoutByNamespace(clientManagedDynamic, common.GvrPod, metav1.ListOptions{}, "gatekeeper-system", 6, true, 240)
 		})
+
+		AfterAll(func() {
+			if CurrentSpecReport().Failed() {
+				utils.KubectlWithOutput("-n", "gatekeeper-system", "get", "pods", "--kubeconfig="+kubeconfigManaged)
```
dhaiducek, line 71: Side note: We should probably put that Gatekeeper namespace in a variable in another PR.

dhaiducek (dismissed):
LGTM!

dhaiducek (approved):
...or this PR. That's cool too. 😄

Makefile
```diff
@@ -29,6 +29,13 @@ KIND_MANAGED_NAMESPACE ?= open-cluster-management-agent-addon
 MANAGED_CLUSTER_NAME ?= managed
 HUB_CLUSTER_NAME ?= hub
 KIND_VERSION ?= latest
+
+# TODO: Remove this since this a workaround until the following is addressed.
+# https://github.com/stolostron/backlog/issues/25698
+ifeq ($(KIND_VERSION), latest)
+	KIND_VERSION = v1.24.4
+endif
```
JustinKuli, line 37: I wish we didn't have to do it like this, but it would be more of a hassle to be "honest" and set the default to v1.24.x, since we would have to update the prow config to require that, instead of latest...

JustinKuli (approved):

Makefile
mprahl: Right, and we'd also have to update every single controller to use `v1.24.x` instead of `latest`.

PR #475, opened 2022-09-16
Kyverno test case failure in canary check wrong cluster

gparvin:
The canary test failed because it's looking for a non-compliant policy on the managed cluster but the kyverno install policy only deploys itself to the hub.  It has an embedded helm chart which goes to the managed cluster.  Changed the namespace check.

Refs:
 - https://github.com/stolostron/backlog/issues/25798


dhaiducek (approved):

dhaiducek:
I went back to double check since I was the one that broke the test. Thanks for the update! I definitely got confused on the Policy `PlacementRule` vs the `PlacementRule` in the Application subscription!

PR #483, opened 2022-09-30
Set `RELEASE_BRANCH` in E2E

dhaiducek:
If this is not set, it'll default to `main` when it's run on release branches.

What I'm not sure on is when this is a `push`. I'm assuming GitHub will return an empty string and the workflow will proceed with `main`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,12 +315,12 @@ e2e-debug: e2e-debug-hub e2e-debug-managed
 .PHONY: e2e-debug-hub
 e2e-debug-hub:
 	mkdir -p $(DEBUG_DIR)
-	kubectl get namespaces --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME) > $(DEBUG_DIR)/hub_get_namespaces.log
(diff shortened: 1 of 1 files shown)


JustinKuli (dismissed):
Seems good. I guess we'll find out what it does on push after this is merged? 😄

dhaiducek:
I think this latest update should work. This way, if it's a push to `governance-policy-framework`, it'll use the base branch name and set up the environment as expected (rather than defaulting to `main`).

With this, I also removed the separate `governance-policy-framework` checkout since it seemed like this new logic made it unnecessary?

.github/workflows/kind.yml
```diff
@@ -44,22 +47,11 @@ jobs:
...
-
-    - name: Checkout Policy Framework
-      # Checkout like "usual" when called from this repository
-      if: ${{ github.event.repository.name == 'governance-policy-framework' }}
-      uses: actions/checkout@v2
-      with:
-        # `repository` is inferred as the "caller" repository 
-        path: framework
-        # `ref` is inferred as the new commit
+        ref: ${{ env.RELEASE_BRANCH }} # like main or release-*
```
JustinKuli, line 54: Let's say I have a PR in this repo that changes some tests. I want this workflow to test my changes... does it do that? It looks like it would just run the old tests - am I missing something?

dhaiducek: Ah, I see what you mean. That makes sense--I'll revert it.

JustinKuli (approved):

PR #489, opened 2022-10-10
Simplify the Gatekeeper tests

mprahl:
Testing the Gatekeeper functionality is out of scope for the BVT/SVT/E2E tests since that is limited to the ACM Governance functionality. This should be handled by the Gatekeeper and Gatekeeper operator tests instead.

This was prompted by these tests often taking longer than the max timeout of 10 minutes that is set, and thus leading to flaky BVT and SVT results.

Relates:
https://github.com/stolostron/backlog/issues/26081

Signed-off-by: mprahl <mprahl@users.noreply.github.com>


ChunxiAlexLuo (approved):
/lgtm

Just testing the gatekeeper operator installation and removing functionality look good to me.

PR #491, opened 2022-10-11
Make `default` the default

dhaiducek:
`make` will run the first target it encounters as a default. Since our linting was included first, the first target there was `yaml-lint`, which was failing.

Alternatively, we could remove the argument-less `make` command, but I wasn't sure of the effect of that.


JustinKuli (approved):

ChunxiAlexLuo (approved):
/lgtm

Thanks for fixing this

PR #492, opened 2022-10-11
Revert Hibernation

dhaiducek:
It seems there's something happening to cause the E2E user to be logged out mid-test. Until that can be resolved, we should probably keep the cluster running.


ChunxiAlexLuo (approved):
/lgtm

PR #506, opened 2022-10-18
Increase maximum test time

JustinKuli:
Previously, we were somewhat limited because the tests run in Travis, whichmight assume things are broken and abort the tests if nothing is written to stdout in a while. Now, a one-liner will prevent that, so we can increase the maximum timeout.

Refs:
 - https://github.com/stolostron/backlog/issues/26165

Signed-off-by: Justin Kulikauskas <jkulikau@redhat.com>

PR #507, opened 2022-10-18
Increase the timeout for the compliance operator tests

gparvin:
The timeout should not be limited by the travis timeout of 10 minutes. Increasing the timeout for scan completions.

Refs:
 - https://github.com/stolostron/backlog/issues/26164

### DIFF
--- a/test/integration/policy_comp_operator_test.go
+++ b/test/integration/policy_comp_operator_test.go
@@ -19,6 +19,8 @@ import (
 	"github.com/stolostron/governance-policy-framework/test/common"
 )
 
+const ComplianceScanTimeoutSeconds = 900
+
 func isOCP46andAbove() bool {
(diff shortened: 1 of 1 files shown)


test/integration/policy_comp_operator_test.go
```diff
@@ -19,6 +19,8 @@ import (
 	"github.com/stolostron/governance-policy-framework/test/common"
 )
 
+const ComplianceScanTimeoutSeconds = 900
```
JustinKuli, line 22: Great minds think alike? I just opened https://github.com/stolostron/governance-policy-framework/pull/506 to increase the "max" timeout to 15 minutes. I'm fine if we want to keep the timeout for this test separate, although I think it is the main use of the "max" timeout...

gparvin:
This is resolved by the change Justin mentioned.

PR #509, opened 2022-10-19
Add CRD sync checking to the periodic CI script

dhaiducek:
We have CRDs synced to `multiclusterhub-operator` and an action running already that checks the `governance-policy-addon-controller`. This adds a check to make sure both of those are up-to-date/passing.

Note that this check will currently be failing until this PR is merged:
- https://github.com/stolostron/multiclusterhub-operator/pull/885

### DIFF
--- a/build/periodic.sh
+++ b/build/periodic.sh
@@ -111,6 +111,62 @@ checkProwJob() {
 	return $rcode
 }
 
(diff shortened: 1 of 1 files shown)


mprahl (approved):
/hold for blocking PR to merge

PR #514, opened 2022-10-27
Delete the Kyverno webhooks before the deployment 

mprahl:
This avoids a race condition where the webhooks are still present but Kyverno is down and blocks API actions. This also fixes a typo in one of the Kyverno webhook deletion commands.


dhaiducek (approved):

dhaiducek (dismissed):

mprahl:
@dhaiducek I added an additional commit to try to further address the flakiness.

ChunxiAlexLuo (approved):
/lgtm

dhaiducek:
That's quite a few deletions to be running through an `Eventually` process. What's the flakiness being addressed, and why would redoing deletions be the solution?

PR #516, opened 2022-10-28
Add missing assertions on Eventually calls

mprahl:
Gomega seems to ignore the Eventually calls if there is no assertion on them.


JustinKuli (approved):
My bad on some of these, thanks for catching it!

dhaiducek (approved):
Good to know! That's definitely how it's documented too: https://onsi.github.io/gomega/#eventually
I should more thoroughly read those docs...eventually 😉

PR #528, opened 2022-11-22
Use an older API version of ManagedClusterSet

mprahl:
This ensures more compatibility across ACM versions.

Relates:
https://github.com/stolostron/backlog/issues/26834

PR #529, opened 2022-11-22
Clean up any stale identities

dhaiducek:
The expectation is that identities would have the pattern `<username>:<username>` because we expect the username to match the IDP. But in case there was a stale IDP, and it's somehow used we want to make sure to clean it up also.


mprahl (approved):

mprahl:
This will depend on #528.

mprahl (approved):

PR #532, opened 2022-12-08
Handle when repos can't be cloned

dhaiducek:
The CICD GitHub token was invalid, so our periodic job couldn't clone the repos. I thought about handling the single false positive to let it all pass, but I think it'd be better that we knew about when the clone doesn't succeed and short circuit the checks.

### DIFF
--- a/build/periodic.sh
+++ b/build/periodic.sh
@@ -14,7 +14,7 @@ cloneRepos() {
 	for prereqrepo in ${UTIL_REPOS}; do
 		if [ ! -d ${prereqrepo} ]; then
 			echo "Cloning ${prereqrepo} ..."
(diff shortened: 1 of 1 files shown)


JustinKuli (commented):
Are there other possible failing commands that we want to fail the job? Ie, should we just add `set -euo pipefail` to the top of this file?

dhaiducek:
> Are there other possible failing commands that we want to fail the job? Ie, should we just add `set -euo pipefail` to the top of this file?

Hmmm. I keep going back and forth on it. I suppose we could, but seeing the run without a token, it seems like other error cases might be handled fairly well and this way we'd still get some information out of it even if there is an error or two midway. On the other hand, we wouldn't know about an error that occurs consistently in the run.

JustinKuli (approved):

PR #533, opened 2022-12-20
Test that the required default metrics are present

mprahl:
Note that the spec-sync metrics are skipped due to it not being available on a self-managed Hub. The presence of the other sync metrics indicate that the metrics are exported from governance-policy-framework addon properly. Other metrics that require error conditions to show up are also skipped.

Relates:
https://issues.redhat.com/browse/ACM-2452

### DIFF
--- a/test/integration/verify_metrics_test.go
+++ b/test/integration/verify_metrics_test.go
@@ -1,0 +1,152 @@
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	"context"
+	"crypto/tls"
+	"encoding/base64"
+	"encoding/json"
+	"io"
+	"net/http"
+	"time"
+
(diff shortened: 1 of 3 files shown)


JustinKuli (approved):

mprahl:
@JustinKuli and @dhaiducek this was updated to use a service account token instead.

dhaiducek:
> @JustinKuli and @dhaiducek this was updated to use a service account token instead.

Sounds good. Once the global sub-admin user is established, we should probably consider changing this flow to use that token instead.

dhaiducek (approved):
LGTM!

The SA/Secret generation could be swapped with the Prometheus URL gathering if the Secret takes longer than a moment so that other tasks are done in the meantime, but I'm not sure doing that really saves all that much time compared to the benefit of having them grouped logically together.

PR #534, opened 2022-12-22
Add an eventually around metrics verification

mprahl:
This is to account for Prometheus scrape times.


dhaiducek (approved):
Thanks for the update, @mprahl!

dhaiducek:
We might have to override to get this in (unless the original code passes in some cases) since the latest snapshot isn't in Quay and the BVT is failing so there won't be a newer snapshot...
https://coreos.slack.com/archives/CVAB9H3DE/p1672793819668599

PR #535, opened 2023-01-05
Fix E2E test for new enum requirement

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-2148


mprahl (approved):

JustinKuli (approved):
Yeah, I think I was just hoping to prevent this dependency from accidentally being satisfied. But this seems fine

PR #536, opened 2023-01-09
Add check for v4.12+ managed cluster for PSP test

dhaiducek:
This skips the PSP test when run against an OCP v4.12 cluster since the PSP is removed in OCP v4.12 and above.

ref: https://issues.redhat.com/browse/ACM-2679


test/common/common.go
```diff
@@ -212,3 +216,31 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+
+// IsMinimumVersion detects OCP versions given an x.y version lower bound
+func IsMinimumVersion(minVersion string) bool {
+	clusterVersion, err := ClientManagedDynamic.Resource(GvrClusterVersion).Get(
+		context.TODO(),
+		"version",
+		metav1.GetOptions{},
+	)
+
+	if err != nil && k8serrors.IsNotFound(err) {
```
mprahl, line 228: What if the error is something else than not found?

test/common/common.go
```diff
@@ -212,3 +216,31 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+		// no version CR, not ocp
+		klog.V(5).Infof("This is not an OCP cluster")
+
+		return false
+	}
+
+	desired := clusterVersion.Object["status"].(map[string]interface{})["desired"]
+	version, _ := desired.(map[string]interface{})["version"].(string)
```
mprahl, line 236: Could you use `unstructured.NestedString` instead?

test/common/common.go
```diff
@@ -212,3 +216,31 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+
+	klog.V(5).Infof(fmt.Sprintf("OCP Version " + version))
+
+	// Convert to valid semantic versions by adding the "v" prefix
+	minSemVer := fmt.Sprintf("v%s", minVersion)
+	ocpSemVer := semver.MajorMinor(fmt.Sprintf("v%s", version))
+
+	// Compare returns: 0 if ver1 == ver2, -1 if ver1 < ver2, or +1 if ver1 > ver2
+	return semver.Compare(ocpSemVer, minSemVer) != -1
```
JustinKuli, line 245: Nit: it might be more usual to write this as `>=0`. In other places I've seen with similar `Compare` functions, the value isn't always restricted to `[-1, 0, 1]`. But it's also possible that `!= -1` is preferred by this library, so in that case just let me know!

test/common/common.go
```diff
@@ -212,3 +216,31 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
 		_, _ = utils.KubectlWithOutput("get", resource, "--all-namespaces", "-o", "yaml", "--kubeconfig="+kubeconfig)
 	}
 }
+
+// IsMinimumVersion detects OCP versions given an x.y version lower bound
+func IsMinimumVersion(minVersion string) bool {
```
JustinKuli, line 221: Nit: `IsAtLeastVersion` or `IsVersionOrAbove` might be more descriptive - at first glance I might guess this function only checked for a specific version.

test/common/common.go
dhaiducek: In case the library changes, that does sound more resilient. I'll make the update!

test/common/common.go
```diff
@@ -212,3 +217,33 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+		context.TODO(),
+		"version",
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// no version CR, not ocp
+			klog.V(5).Infof("This is not an OCP cluster")
+		} else {
+			klog.V(5).Infof("Encountered an error fetching the OCP version: %v", err)
```
mprahl, line 233:
```suggestion
			klog.Infof("Encountered an error fetching the OCP version: %v", err)
```

test/common/common.go
```diff
@@ -212,3 +217,33 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+		}
+
+		return false
+	}
+
+	version, _, _ := unstructured.NestedString(clusterVersion.Object, "status", "desired", "version")
+
+	klog.V(5).Infof(fmt.Sprintf("OCP Version " + version))
```
mprahl, line 241:
```suggestion
	klog.V(5).Info("OCP Version " + version)
```

test/common/common.go
```diff
@@ -212,3 +217,33 @@ func OutputDebugInfo(testName string, kubeconfig string, additionalResources ...
...
+func IsAtLeastVersion(minVersion string) bool {
+	clusterVersion, err := ClientManagedDynamic.Resource(GvrClusterVersion).Get(
+		context.TODO(),
+		"version",
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			// no version CR, not ocp
+			klog.V(5).Infof("This is not an OCP cluster")
```
mprahl, line 231:
```suggestion
			klog.V(5).Info("This is not an OCP cluster")
```

JustinKuli (approved):
/retest
/lgtm

dhaiducek:
For reference, the ClusterPool has been suffering from a lack of resources in AWS in a particular sub-region. I'll dig into it a bit more if this one doesn't provision.

dhaiducek:
I think Hive might default to `us-east-1c`. I'll see if there's an option to unset it or set it to something else.
```
time="2023-01-11T16:57:09Z"
level=error
msg="Error: Error launching source instance: InsufficientInstanceCapacity: We currently do not have sufficient m6a.2xlarge capacity in the Availability Zone you requested (us-east-1c). Our system will be working on provisioning additional capacity. You can currently get m6a.2xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1a, us-east-1b, us-east-1d, us-east-1f."
time="2023-01-11T16:57:09Z" level=error msg="\tstatus code: 500, request id: 698f9542-c0bf-4313-bb9d-4be62ee834cb"
```

dhaiducek:
/retest

I had set the install config to use `us-east-1c`, so hopefully setting it to `us-east-1f` will resolve it. (I forgot that setting the zone would allow a single NAT gateway by preventing AWS from deploying to multiple zones, which would require multiple NAT gateways and additional cost.)

PR #541, opened 2023-01-19
Add test for configpolicy cleanup during uninstall

JustinKuli:
Tests that the ConfigurationPolicy CRD does not become stuck due to pruneObjectBehavior finalizers when the controller is uninstalled.

Refs:
 - https://issues.redhat.com/browse/ACM-2923


JustinKuli:
The tests here should fail until https://github.com/open-cluster-management-io/config-policy-controller/pull/96 is merged into stolostron.

But this previous run shows it passing with a dev image: https://github.com/stolostron/governance-policy-framework/actions/runs/3969435606/jobs/6803925704, so I think it's ready for initial review at least.

mprahl (dismissed):

mprahl (approved):

PR #542, opened 2023-01-20
Remove the unused build harness

mprahl:
This would often cause CI failures when the downloading would fail. The build harness doesn't actually seem to be used, so there's no reason to download it.


dhaiducek:
`.travis.yml` needs to be reworked for this to happen, but I've been thinking this is some technical debt we should address.

mprahl:
@dhaiducek I just updated the travis.yml file. What do you think?

dhaiducek (approved):
LGTM! I can't think of any reason not to merge this in!

PR #554, opened 2023-02-07
add lastest msg check

yiraeChristineKim:
Signed-off-by: Yi Rae Kim <yikim@redhat.com>

from ACM-3075
GRC Policies have tests that check for a compliance or noncompliance and then some action is taken so that the compliance state changes.  It would be useful after one of these tests, to take a look at the compliance history and see if it was updated as expected. 

This has been proposed as the result of sprint retrospective discussion.

### DIFF
--- a/test/common/policy_utils.go
+++ b/test/common/policy_utils.go
@@ -4,6 +4,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"path/filepath"
 	"strings"
 
@@ -166,29 +167,36 @@ func DoRootComplianceTest(policyName string, compliance policiesv1.ComplianceSta
 	).Should(Equal(compliance))
 }
 
+func GetHistoryMessages(policyName string, templateIdx int) ([]interface{}, bool, error) {
+	empty := make([]interface{}, 0)
+	replicatedPolicyName := UserNamespace + "." + policyName
+	policyInterface := ClientManagedDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
+
+	policy, err := policyInterface.Get(context.TODO(), replicatedPolicyName, metav1.GetOptions{})
+	if err != nil {
+		return empty, false, errors.New("error in getting policy")
+	}
+
+	details, found, err := unstructured.NestedSlice(policy.Object, "status", "details")
+	if !found || err != nil || len(details) <= templateIdx {
+		return empty, false, errors.New("error in getting status")
+	}
+
+	templateDetails, ok := details[templateIdx].(map[string]interface{})
+	if !ok {
+		return empty, false, errors.New("error in getting detail")
+	}
+
+	history, found, err := unstructured.NestedSlice(templateDetails, "history")
+
+	return history, found, err
+}
+
 // GetLatestStatusMessage returns the most recent status message for the given policy template.
 // If the policy, template, or status do not exist for any reason, an empty string is returned.
 func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
 	return func() string {
-		replicatedPolicyName := UserNamespace + "." + policyName
-		policyInterface := ClientManagedDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
-
-		policy, err := policyInterface.Get(context.TODO(), replicatedPolicyName, metav1.GetOptions{})
-		if err != nil {
-			return ""
-		}
-
-		details, found, err := unstructured.NestedSlice(policy.Object, "status", "details")
-		if !found || err != nil || len(details) <= templateIdx {
-			return ""
-		}
-
-		templateDetails, ok := details[templateIdx].(map[string]interface{})
-		if !ok {
-			return ""
-		}
-
-		history, found, err := unstructured.NestedSlice(templateDetails, "history")
+		history, found, err := GetHistoryMessages(policyName, templateIdx)
 		if !found || err != nil || len(history) == 0 {
 			return ""
 		}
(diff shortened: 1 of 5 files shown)


test/common/policy_utils.go
```diff
@@ -204,6 +205,15 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
 	}
 }
 
+func DoHistoryUpdatedTest(policyName string, compliance policiesv1.ComplianceState, templateIdx int) {
```
mprahl, line 208: Could you please change this so that you are checking the whole history?

Part of the goal of this test is to ensure that not only that the latest status message is correct but that all the states we expected have occurred.

mprahl: Could you please also check the status message in each entry of the history?

mprahl:
@yiraeChristineKim could you please conform your commit messages and structure to our guidelines?
https://github.com/open-cluster-management-io/community/blob/main/sig-policy/contribution-guidelines.md#commit-messages
https://github.com/open-cluster-management-io/community/blob/main/sig-policy/contribution-guidelines.md#commit-structure

test/common/policy_utils.go
```diff
@@ -166,29 +168,36 @@ func DoRootComplianceTest(policyName string, compliance policiesv1.ComplianceSta
...
+	replicatedPolicyName := UserNamespace + "." + policyName
+	policyInterface := ClientManagedDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
 
-		policy, err := policyInterface.Get(context.TODO(), replicatedPolicyName, metav1.GetOptions{})
-		if err != nil {
-			return ""
-		}
+	policy, err := policyInterface.Get(context.TODO(), replicatedPolicyName, metav1.GetOptions{})
+	if err != nil {
+		return empty, false, errors.New("error in getting policy")
```
JustinKuli, line 178: Nitpick: instead of `errors.New`, this could wrap the error like `fmt.Errorf("failed to get policy for history: %w", err)`. This would give us a little more information if it fails during a test.

test/common/policy_utils.go
```diff
@@ -204,6 +213,25 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
...
+			By("Getting history message from history array")
+			historyItem, ok := history[i].(map[string]interface{})
+			g.Expect(ok).Should(BeTrue())
+			order := i + 1
+			By("Checking message is same as [" + msg + "]" + " order number is " + fmt.Sprintf("%d", order))
+			message, _, _ := unstructured.NestedString(historyItem, "message")
+			fmt.Println(msg)
+			fmt.Println(message)
+			g.Expect(message).Should(Equal(msg))
+		}
```
JustinKuli, line 231: I think if the number of messages in the history doesn't match the wanted number of messages, this could run into some trouble. Something might panic, or some messages won't get checked.

Instead of checking them one by one, if this turned the `history` into a simple list of messages, then I think the check could just be something like `g.Expect(historyMessages).Should(Equal(messages))`. And if Gomega does the comparison itself, then I think it will give better debugging information about what items don't match when it fails.

test/e2e/cert_policy_test.go
```diff
@@ -179,6 +178,26 @@ var _ = Describe("Test cert policy", func() {
...
+					".*.test.com Disallowed: [\\*]: default:cert-policy-secret",
+				"Compliant",
+				"NonCompliant;  1 CA certificates exceed the maximum duration of 26280h0m0s: "+
+					"default:cert-policy-secret",
+				"Compliant",
+				"NonCompliant;  1 certificates exceed the maximum duration of 9528h0m0s: "+
+					"default:cert-policy-secret",
+				"Compliant",
+				"NonCompliant;  1 CA certificates expire in less than 45h0m0s: default:cert-policy-secret",
+			)
```
JustinKuli, line 197: Since this is a test on its own, I think it should be in an `It`, not inside the `AfterAll`. Since this `Describe` block has the `Ordered` attribute, the `It` statements will always be done in the right order. `AfterAll` should just be for cleanup that always needs to take place, even when a test fails.

JustinKuli (changes requested):
The bulk of this looks really good! I like the way `GetLatestStatusMessage` was split up to get the history messages when necessary.

The only change I think is really necessary is moving those tests into separate `It` blocks; I want to keep the `AfterAll` sections clean.

test/common/policy_utils.go
yiraeChristineKim: how we know which history message broken?

test/e2e/cert_policy_test.go
yiraeChristineKim: ok!

test/e2e/configuration_policy_test.go
```diff
@@ -180,10 +205,37 @@ var _ = Describe("Test configuration policy", Ordered, func() {
...
+					"in namespaces: default",
+				"Compliant; notification - roles [role-policy-e2e] in namespace default "+
+					"was missing, and was created successfully",
+				"NonCompliant; violation - No instances of `roles` found as specified "+
+					"in namespaces: default",
+				"NonCompliant; violation - roles not found: [role-policy-e2e] in "+
+					"namespace default missing",
+			)
+		})
 		AfterAll(func() {
```
yiraeChristineKim, line 229: these clean up very important... without this... this test inconsistent

test/e2e/configuration_policy_test.go
```diff
@@ -218,9 +270,26 @@ var _ = Describe("Test configuration policy", Ordered, func() {
...
+				"NonCompliant; violation - roles found: [role-policy-e2e] in namespace default",
+				"Compliant; notification - roles [role-policy-e2e] in namespace default missing "+
+					"as expected, therefore this Object template is compliant",
+			)
+		})
 		AfterAll(func() {
+			By("Deleting the role, policy, and events on managed cluster")
 			common.DoCleanupPolicy(rolePolicyYaml, common.GvrConfigurationPolicy)
 			_, err := common.OcManaged(
+				"delete", "events", "-n", "managed",
```
yiraeChristineKim, line 287: without this, these events affect other tests so next tests are failed

test/e2e/iam_policy_test.go
```diff
@@ -36,8 +36,34 @@ var _ = Describe("Test iam policy", func() {
...
 		AfterAll(func() {
+			By("Deleting the policy and events on managed cluster")
 			common.DoCleanupPolicy(iamPolicyYaml, common.GvrIamPolicy)
+			_, err := common.OcManaged(
+				"delete", "-f", "../resources/iam_policy/clusterrolebinding.yaml",
+				"--ignore-not-found",
+			)
+			Expect(err).To(BeNil())
+			_, err = common.OcManaged(
+				"delete", "clusterrolebinding",
```
yiraeChristineKim, line 56: sometimes it doesn't remove at above, I guess because of events.. so I clean up here one more time

JustinKuli (dismissed):
LGTM! Thanks

mprahl:
@yiraeChristineKim could you please squash your commits to conform to our standards?
https://github.com/open-cluster-management-io/community/blob/main/sig-policy/contribution-guidelines.md#commit-messages
https://github.com/open-cluster-management-io/community/blob/main/sig-policy/contribution-guidelines.md#commit-structure

JustinKuli (approved):
/unhold

PR #556, opened 2023-02-09
Use a common GitOps user for generator tests

JustinKuli:
This should reduce flakiness caused by the auth pods restarting each time we create a new gitops user.

Refs:
 - https://issues.redhat.com/browse/ACM-2423

### DIFF
--- a/test/common/common.go
+++ b/test/common/common.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	"golang.org/x/mod/semver"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


test/common/common.go
```diff
@@ -247,3 +248,31 @@ func IsAtLeastVersion(minVersion string) bool {
...
+					context.TODO(), namespace, metav1.GetOptions{},
+				)
+				isNotFound := k8serrors.IsNotFound(err)
+				if !isNotFound && err != nil {
+					GinkgoWriter.Printf("'%s' namespace 'get' error: %w", err)
+				}
+
+				return isNotFound
+			},
+			DefaultTimeoutSeconds*2,
```
dhaiducek, line 274: Since this has been a flake before, should we go ahead and make this longer?
```suggestion
			DefaultTimeoutSeconds*4,
```

dhaiducek (approved):
/hold

LGTM! Holding in case we want to extend the namespace deletion timeout.

dhaiducek (approved):
Holy moly! This is working! Sorry, I totally missed that it passed! Nice work!

test/common/gitops_utils.go
```diff
@@ -72,15 +103,21 @@ func GitOpsUserSetup(
 	}
 
 	By("Cleaning up any existing subscription-admin user config")
-	GitOpsCleanup(namespace, ocpUser)
+	GitOpsCleanup(ocpUser)
+
+	for _, ns := range gitopsTestNamespaces {
+		CleanupHubNamespace(ns)()
```
dhaiducek, line 109: Wait a min. What's up with that extra set of parentheses?

dhaiducek: Oh wild. I was wondering about this when I saw the `Eventually`. So it's returning a function so that it uses the Gomega that's currently in action? Why not just pass the Gomega object in instead of returning an entire function (assuming that's the reason)?

JustinKuli: It's a function that returns a function. The most common use of the function is:
```go
AfterAll(CleanupHubNamespace("foobar"))
```
because `AfterAll` takes a function.

It could be written to be less weird here, but then the more usual case would look like:
```go
AfterAll(func() {
    CleanupHubNamespace("foobar")
})

dhaiducek: I do like how much cleaner it is, but I think in the end it's confusing since it won't look like the other AfterAlls() and if I want to use `CleanupHubNamespace()` alongside other functions, I'd need to use the funky syntax.

JustinKuli: Good point. Updated 👍

dhaiducek (approved):

PR #557, opened 2023-02-09
Ensure the iam group is cleaned up

JustinKuli:
In one test run, the group was left behind, which made future test runs fail earlier, and it would never recover automatically.


dhaiducek (dismissed):

dhaiducek:
@JustinKuli A rebase on this one should get the CI to pass since the Gatekeeper tests were removed, assuming the PR is still relevant.

dhaiducek (approved):

PR #563, opened 2023-02-15
Remove the upstream gatekeeper operator test

gparvin:
The upstream gatekeeper operator test uses an old operator version that is not longer supported and does not work on newer k8s versions including OCP 4.12.

Refs:
 - https://issues.redhat.com/browse/ACM-3320


gparvin:
/hold
I created this PR quickly and need to double check whether there are additional things to remove.

gparvin:
/unhold
Please take a look now and provide review feedback. Thanks

gparvin:
I don't think anything uses https://github.com/stolostron/governance-policy-framework/blob/main/deploy/gatekeeper/gatekeeper.yaml and I can't tell why we have it.  Anyone know?  I'm going to see what happens when I remove it

dhaiducek:
> I don't think anything uses https://github.com/stolostron/governance-policy-framework/blob/main/deploy/gatekeeper/gatekeeper.yaml and I can't tell why we have it. Anyone know? I'm going to see what happens when I remove it

Looks like it was added along with `gatekeeper_test.go`, so without further investigation, to me it looks safe to remove.

JustinKuli:
More links:
- https://github.com/stolostron/backlog/issues/27188
- https://github.com/stolostron/backlog/issues/27182

I had briefly looked at making a similar PR, but you're ahead of me :D

gparvin:
The failed test is `policy_imagemanifest_test.go:20` which I think is unrelated.

test/integration/policy_gatekeeper_operator_downstream_test.go
```diff
@@ -10,13 +10,36 @@ import (
...
+	version, _ := desired.(map[string]interface{})["version"].(string)
+
+	klog.V(5).Infof("OCP Version %s\n", version)
+
+	return strings.HasPrefix(version, "4.4")
+}
+
 var _ = Describe("RHACM4K-3055", Ordered, Label("policy-collection", "stable", "BVT"), func() {
 	BeforeAll(func() {
 		if isOCP44() {
```
dhaiducek, line 45: Would you please use `IsAtLeastVersion` instead and remove the `isOCP44` function?

Alternatively, it seems like this could be removed all together since we're well beyond `4.4` at this point.

dhaiducek (changes requested):

test/integration/policy_gatekeeper_operator_downstream_test.go
gparvin: I went with removing it.  Good point -- I didn't notice it was a 4.4 specific check.  Thanks

dhaiducek (approved):

PR #572, opened 2023-03-01
Add basic tests for ACM Policies in hosted mode in the policy framework E2E tests

yiraeChristineKim:
Add basic tests for ACM Policies in hosted mode in the policy framework E2E tests
We currently test that hosted mode deployments of the policy controllers work as expected on PRs to the Policy Addon controller. This is good for testing changes in how the controllers are deployed but it doesn't test their functionality in hosted mode and it also doesn't gate code changes the policy controllers.

GRC currently has E2E and other integration tests in the stolostron/governance-policy-framework repo that gate code changes from being fast-forwarded to a release branch and thus getting into a release build. A new E2E test should be added that tests the policy controllers in hosted mode.

Since these E2E tests are already executed on PR CI in the Stolostron repos before an upstream PR is synced, this will act as a gate for controller changes that break hosted mode deployments.
NOTE: 
- hosted mode takes more time than normal mode
- some Mac has issue on Managedclusteraddon
Refs:

https://issues.redhat.com/browse/ACM-3076
Signed-off-by: Yi Rae Kim [yikim@redhat.com](mailto:yikim@redhat.com)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -11,6 +11,10 @@ on:
       - release-2.[7-9]
 
 jobs:
+  kind-tests-hosted-mode:
+    name: Framework KinD hosted mode Workflow
+    uses: ./.github/workflows/kind_host.yml
+  
   kind-tests:
     name: Framework KinD # Part of the check name, be careful when changing.
     uses: ./.github/workflows/kind.yml
--- a/.github/workflows/kind_host.yml
+++ b/.github/workflows/kind_host.yml
@@ -1,0 +1,138 @@
+name: Framework KinD Hosted mode Workflow
+
+on:
+  workflow_call:
+    inputs: # To specify where they're located for components, it's leveraged by our controller repos.
+      hub_only_component:
+        required: false
+        type: string
+        default: 'false'
+
+env:
+  RELEASE_BRANCH: ${{ github.event.pull_request.base.ref || github.ref_name }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  tests:
+    name: Hosted mode Tests # Part of the check name, be careful when changing.
+
+    runs-on: ubuntu-latest
+    env:
+      REGISTRY: localhost:5000
+
+    steps:
+
+    - name: Checkout Component Repository
+      # Checkout a specific component when called from another repository
+      if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+      uses: actions/checkout@v3
+      with:
+        # `repository` is inferred as the "caller" repository 
+        path: component
+        # `ref` is inferred as the new commit in the "caller" repository
+
+    - name: Checkout Policy Framework
+      # Checkout a stable branch when called from another repository
+      if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+      uses: actions/checkout@v3
+      with:
+        repository: stolostron/governance-policy-framework
+        path: framework
+        ref: ${{ env.RELEASE_BRANCH }} # like main or release-*
+
+    - name: Checkout Policy Framework
+      # Checkout like "usual" when called from this repository
+      if: ${{ github.event.repository.name == 'governance-policy-framework' }}
+      uses: actions/checkout@v3
+      with:
+        # `repository` is inferred as the "caller" repository 
+        path: framework
+        # `ref` is inferred as the new commit
+
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: framework/go.mod
+
+    - name: Bootstrap the KinD clusters for hosted mode
+      working-directory: framework
+      run: |
+        echo "::group::make e2e-dependencies"
+        make e2e-dependencies
+        echo "::endgroup::"
+
+        echo "::group::make setup hosted mode"
+        make kind-bootstrap-hosted
+        echo "::endgroup::"
+        echo "::group::Saving kubeconfig paths for use in other steps"
+    
+        echo "MANAGED_KUBECONFIG=$(pwd)/kubeconfig_managed" >> $GITHUB_ENV
+        echo "HUB_KUBECONFIG=$(pwd)/kubeconfig_hub" >> $GITHUB_ENV
+        echo "HUB_INTERNAL_KUBECONFIG=$(pwd)/kubeconfig_hub_internal" >> $GITHUB_ENV
+        echo "::endgroup::"
+    
+        echo "::group::test for patch step"
+        export MANAGED_CONFIG=$(pwd)/kubeconfig_managed
+        export HUB_CONFIG=$(pwd)/kubeconfig_hub
+        export HUB_CONFIG_INTERNAL=$(pwd)/kubeconfig_hub_internal
+        echo "::endgroup::"
+
+    - name: Patch Component Image
+      if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+      working-directory: component
+      env:
+        WATCH_NAMESPACE: cluster2-hosted
+      run: |
+        if [ "$RELEASE_BRANCH" != main ]; then
+          export VERSION_TAG=latest-$(echo $RELEASE_BRANCH | cut -d'-' -f 2):
+        else
+          export VERSION_TAG=latest
+        fi
+
+        echo "Use $VERSION_TAG"
+
+        echo " KIND_NAME=policy-addon-ctrl1 and the hub config as the default kubeconfig"
+
+        echo "Using KIND_NAME=policy-addon-ctrl1 and the hub config as the default kubeconfig"
+        export KIND_NAME=policy-addon-ctrl1
+        export KUBECONFIG=${HUB_KUBECONFIG}
+
+        echo "::group::make build-images"
+        TAG=$(VERSION_TAG) make build-images
+        echo "::endgroup::"
+
+        echo "::group::make kind-deploy-controller-dev"
+        TAG=$(VERSION_TAG) KIND_NAMESPACE=cluster2-hosted make kind-deploy-controller-dev
+        echo "::endgroup::"
+
+    - name: Run e2e tests
+      working-directory: framework
+      env:
+        TEST_ARGS: --json-report=report.json --junit-report=report.xml --output-dir=test-output
+      run: |
+        make e2e-test-hosted
+
+    - name: Debug
+      if: ${{ failure() }}
+      working-directory: framework
+      run: |
+        make e2e-debug-kind
+        tar -czvf e2e-debug.tar.gz test-output/debug
+    
+    - name: Upload Debug Artifacts
+      if: ${{ failure() }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: fw-kind-debug-hosted-mode
+        path: framework/e2e-debug.tar.gz
+
+    - name: Upload Test Reports
+      uses: actions/upload-artifact@v3
+      with:
+        name: fw-kind-report-hosted-mode
+        path: |
+          framework/test-output/report.xml
+          framework/test-output/report.json
--- a/Makefile
+++ b/Makefile
@@ -233,13 +233,23 @@ install-resources:
 e2e-dependencies:
 
 K8SCLIENT ?= oc
+GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+MANAGED_CLUSTER_NAMESPACE ?= managed
+
 .PHONY: e2e-test
 e2e-test:
 	@if [ -z "$(TEST_FILE)" ]; then\
-		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
+		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAMESPACE) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB);\
 	else\
-		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
+		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAMESPACE) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB);\
 	fi
+
+.PHONY: e2e-test-hosted
+e2e-test-hosted: CLUSTER_NAMESPACE_ON_HUB=cluster2 
+e2e-test-hosted: IS_HOSTED=true 
+e2e-test-hosted: MANAGED_CLUSTER_NAMESPACE=cluster2-hosted
+e2e-test-hosted: e2e-test
 
 .PHONY: e2e-debug
 e2e-debug: e2e-debug-hub e2e-debug-managed
@@ -309,3 +319,42 @@ integration-test:
 	else\
 		$(GINKGO) -v $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/integration;\
 	fi
+
+#hosted
+ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
+
+.PHONY: kind-bootstrap-hosted
+kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-deploy-cert-manager setup-managedcluster
+
+.PHONY: kind-install-hosted
+kind-install-hosted: $(ADDON_CONTROLLER)
+	@cd ./governance-policy-addon-controller && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_$(HUB_CLUSTER_NAME)
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_$(HUB_CLUSTER_NAME)_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_$(MANAGED_CLUSTER_NAME)
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_$(MANAGED_CLUSTER_NAME)_internal
+	@echo installing policy-propagator on hub
+	-kubectl create ns $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/governance-policy-propagator/$(RELEASE_BRANCH)/deploy/operator.yaml -n $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+
+$(ADDON_CONTROLLER):
+	git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git
+
+.PHONY: setup-managedcluster
+setup-managedcluster: MANAGED_CLUSTER_NAMESPACE=cluster2-hosted
+setup-managedcluster: 
+	-kubectl create ns $(MANAGED_CLUSTER_NAMESPACE) --kubeconfig=kubeconfig_$(HUB_CLUSTER_NAME)
+	-kubectl -n $(MANAGED_CLUSTER_NAMESPACE) create secret generic config-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=kubeconfig_$(HUB_CLUSTER_NAME)
+	-kubectl -n $(MANAGED_CLUSTER_NAMESPACE) create secret generic cert-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=kubeconfig_$(HUB_CLUSTER_NAME)
+	-kubectl -n $(MANAGED_CLUSTER_NAMESPACE) create secret generic iam-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=kubeconfig_$(HUB_CLUSTER_NAME)
+	-sed 's/imagetag/$(VERSION_TAG)/g' test/resources/hosted_mode/managed-cluster-addon.yaml | kubectl apply -f- --kubeconfig=kubeconfig_$(HUB_CLUSTER_NAME) -n cluster2
+
+kind-delete-hosted: $(ADDON_CONTROLLER)
+	@cd governance-policy-addon-controller && make kind-bootstrap-delete-clusters 
+	@rm kubeconfig_hub kubeconfig_managed kubeconfig_hub_internal kubeconfig_managed_internal
+
+.PHONY: 
+kind-deploy-cert-manager:
+	@echo installing cert-manager on managed
+	kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.5.5/cert-manager.yaml --kubeconfig=$(PWD)/kubeconfig_$(MANAGED_CLUSTER_NAME)
+
--- a/test/common/policy_utils.go
+++ b/test/common/policy_utils.go
@@ -23,7 +23,7 @@ import (
 // GetComplianceState returns a function usable by ginkgo.Eventually that retrieves the
 // compliance state of the input policy in the globally configured managed cluster.
 func GetComplianceState(policyName string) func(Gomega) interface{} {
-	return GetClusterComplianceState(policyName, ClusterNamespace)
+	return GetClusterComplianceState(policyName, ClusterNamespaceOnHub)
 }
 
 // GetClusterComplianceState returns a function usable by ginkgo.Eventually that retrieves the
@@ -60,7 +60,7 @@ func PatchPlacementRule(namespace, name string) error {
 		`-p=[{
 			"op": "replace",
 			"path": "/spec/clusterSelector",
-			"value":{"matchExpressions":[{"key": "name", "operator": "In", "values": ["`+ClusterNamespace+`"]}]}
+			"value":{"matchExpressions":[{"key": "name", "operator": "In", "values": ["`+ClusterNamespaceOnHub+`"]}]}
 		}]`,
 	)
 
@@ -91,11 +91,11 @@ func DoCreatePolicyTest(policyFile string, templateGVRs ...schema.GroupVersionRe
 
 	if ManuallyPatchDecisions {
 		plrName := policyName + "-plr"
-		By("Patching " + plrName + " with decision of cluster " + ClusterNamespace)
+		By("Patching " + plrName + " with decision of cluster " + UserNamespace)
 		plr := utils.GetWithTimeout(
 			ClientHubDynamic, GvrPlacementRule, plrName, UserNamespace, true, DefaultTimeoutSeconds,
 		)
-		plr.Object["status"] = utils.GeneratePlrStatus(ClusterNamespace)
+		plr.Object["status"] = utils.GeneratePlrStatus(ClusterNamespaceOnHub)
 		_, err := ClientHubDynamic.Resource(GvrPlacementRule).Namespace(UserNamespace).UpdateStatus(
 			context.TODO(),
 			plr,
@@ -107,7 +107,7 @@ func DoCreatePolicyTest(policyFile string, templateGVRs ...schema.GroupVersionRe
 	managedPolicyName := UserNamespace + "." + policyName
 	By("Checking " + managedPolicyName + " on managed cluster in ns " + ClusterNamespace)
 	mplc := utils.GetWithTimeout(
-		ClientManagedDynamic, GvrPolicy, managedPolicyName, ClusterNamespace, true, DefaultTimeoutSeconds,
+		ClientHostingDynamic, GvrPolicy, managedPolicyName, ClusterNamespace, true, DefaultTimeoutSeconds,
 	)
 	ExpectWithOffset(1, mplc).NotTo(BeNil())
 
@@ -116,7 +116,7 @@ func DoCreatePolicyTest(policyFile string, templateGVRs ...schema.GroupVersionRe
 		By("Checking that the policy template " + typedName + " is present on the managed cluster")
 
 		tmplPlc := utils.GetWithTimeout(
-			ClientManagedDynamic, tmplGVR, policyName, ClusterNamespace, true, DefaultTimeoutSeconds,
+			ClientHostingDynamic, tmplGVR, policyName, ClusterNamespace, true, DefaultTimeoutSeconds,
 		)
 		ExpectWithOffset(1, tmplPlc).NotTo(BeNil())
 	}
@@ -171,7 +171,7 @@ func DoRootComplianceTest(policyName string, compliance policiesv1.ComplianceSta
 func GetHistoryMessages(policyName string, templateIdx int) ([]interface{}, bool, error) {
 	empty := make([]interface{}, 0)
 	replicatedPolicyName := UserNamespace + "." + policyName
-	policyInterface := ClientManagedDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
+	policyInterface := ClientHostingDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
 
 	policy, err := policyInterface.Get(context.TODO(), replicatedPolicyName, metav1.GetOptions{})
 	if err != nil {
@@ -221,6 +221,7 @@ func DoHistoryUpdatedTest(policyName string, messages ...string) {
 		messages = messages[:10]
 	}
 
+	By("Getting policy history, check latest message")
 	Eventually(func(g Gomega) {
 		history, _, err := GetHistoryMessages(policyName, 0)
 		g.Expect(err).Should(BeNil())
@@ -234,7 +235,8 @@ func DoHistoryUpdatedTest(policyName string, messages ...string) {
 			fmt.Println(fmt.Sprint(i) + ": " + m)
 		}
 		By("Check history length is same")
-		g.Expect(len(history)).Should(Equal(lenMessage))
+		g.Expect(history).Should(HaveLen(lenMessage))
+
 		By("Check history message same")
 		g.Expect(strings.Join(historyMsgs, "")).Should(Equal(strings.Join(messages, "")))
 	}, DefaultTimeoutSeconds, 1).Should(Succeed())
@@ -274,7 +276,7 @@ func setRemediationAction(
 		g.ExpectWithOffset(1, err).To(BeNil())
 	}, DefaultTimeoutSeconds, 1).Should(Succeed())
 
-	managedPolicyClient := ClientManagedDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
+	managedPolicyClient := ClientHostingDynamic.Resource(GvrPolicy).Namespace(ClusterNamespace)
 
 	By("Checking that remediationAction = " + remediationAction + " on replicated policy")
 	EventuallyWithOffset(1, func(g Gomega) {
@@ -291,7 +293,7 @@ func setRemediationAction(
 		typedName := tmplGVR.String() + "/" + policyName
 		By("Checking that remediationAction = " + remediationAction + " on policy template " + typedName)
 
-		templateClient := ClientManagedDynamic.Resource(tmplGVR).Namespace(ClusterNamespace)
+		templateClient := ClientHostingDynamic.Resource(tmplGVR).Namespace(ClusterNamespace)
 
 		EventuallyWithOffset(1, func(g Gomega) {
 			template, err := templateClient.Get(ctx, policyName, metav1.GetOptions{})
--- a/test/configuration_policy_prune.go
+++ b/test/configuration_policy_prune.go
@@ -8,8 +8,10 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/stolostron/governance-policy-framework/test/common"
 	. "github.com/stolostron/governance-policy-framework/test/common"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
 	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
@@ -29,7 +31,7 @@ func ConfigPruneBehavior(labels ...string) bool {
 			Expect(err).To(BeNil())
 
 			Eventually(func(g Gomega) {
-				outManaged, err := OcManaged(
+				outManaged, err := OcHosting(
 					"get", "configurationpolicies", "-A",
 				)
 				GinkgoWriter.Printf("cleanPolicy OcManaged configurationpolicy output: %v\n", outManaged)
@@ -37,14 +39,14 @@ func ConfigPruneBehavior(labels ...string) bool {
 				g.Expect(err).To(BeNil())
 			}, DefaultTimeoutSeconds, 1).Should(Succeed())
 
-			outManaged, err := OcManaged(
+			outManaged, err := OcHosting(
 				"delete", "events", "-n", ClusterNamespace,
 				"--field-selector=involvedObject.name="+UserNamespace+"."+policyName,
 				"--ignore-not-found",
 			)
 			GinkgoWriter.Printf("cleanPolicy OcManaged policy event output: %v\n", outManaged)
 			Expect(err).To(BeNil())
-			outManaged, err = OcManaged(
+			outManaged, err = OcHosting(
 				"delete", "events", "-n", ClusterNamespace,
 				"--field-selector=involvedObject.name="+policyName,
 				"--ignore-not-found",
@@ -57,13 +59,21 @@ func ConfigPruneBehavior(labels ...string) bool {
 	pruneTestCreatedByPolicy := func(policyName, policyYaml string, cmShouldBeDeleted bool) {
 		clientManagedDynamic := NewKubeClientDynamic("", KubeconfigManaged, "")
 
+		var clientHostingDynamic dynamic.Interface
+
+		if common.IsHosted {
+			clientHostingDynamic = NewKubeClientDynamic("", KubeconfigHub, "")
+		} else {
+			clientHostingDynamic = NewKubeClientDynamic("", KubeconfigManaged, "")
+		}
+
 		DoCreatePolicyTest(policyYaml, GvrConfigurationPolicy)
 
 		DoRootComplianceTest(policyName, policiesv1.Compliant)
 
 		By("Checking if the status of ConfigurationPolicy " + policyName + " is Compliant")
 		Eventually(func() string {
-			cfgPol := utils.GetWithTimeout(clientManagedDynamic, GvrConfigurationPolicy,
+			cfgPol := utils.GetWithTimeout(clientHostingDynamic, GvrConfigurationPolicy,
 				policyName, ClusterNamespace, true, DefaultTimeoutSeconds)
 
 			compliant, _, _ := unstructured.NestedString(cfgPol.Object, "status", "compliant")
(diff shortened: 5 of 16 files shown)


Makefile
```diff
@@ -168,10 +168,10 @@ deploy-community-policy-framework-managed: deploy-policy-framework-managed-crd-o
 .PHONY: kind-deploy-policy-framework
 kind-deploy-policy-framework:
 	@echo installing policy-propagator on hub
-	kubectl create ns $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	-kubectl create ns $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
```
yiraeChristineKim, line 171: In the hosted mode, it already exit so I added -

go.mod
```diff
@@ -10,13 +10,27 @@ require (
 	github.com/prometheus/common v0.37.0
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/crypto v0.0.0-20220722155217-630584e8d5aa
-	k8s.io/api v0.23.9
-	k8s.io/apimachinery v0.23.9
+	k8s.io/api v0.24.0
```
yiraeChristineKim, line 13: this version cause no conflict

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,171 @@
+name: Framework KinD hosted mode Workflow
+
+on:
+  workflow_call:
+    inputs:
+      path: 
+        required: false
```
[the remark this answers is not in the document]
pshickeydev: Does it error if you don't define the input properties?

So just
```
on:
  workflow_call:

```
I don't see anything in the actions documentation that says `inputs` is required.

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,157 @@
...
+    strategy:
+      # Don't skip all tests if one fails - others may still have useful info
+      fail-fast: false
+      matrix:
+        # Run tests against a new version of kubernetes
+        # KinD tags are listed at https://hub.docker.com/r/kindest/node/tags
+        kind:
+          - 'latest'
+        deployOnHub:
+          - 'false'
```
mprahl, line 29: This isn't applicable to this setup so could you please remove this from the matrix?

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,157 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: framework/go.mod
+
+    - name: Verify modules
+      if: ${{ github.event.repository.name == 'governance-policy-framework' }}
+      working-directory: framework
+      run: |
+        go mod verify
```
mprahl, line 68: This can be removed since that is already tested in the other test.

go.mod
mprahl: What was the conflict?

Makefile
```diff
@@ -349,3 +349,27 @@ integration-test:
...
 	fi
+
+#hosted
+
+.PHONY: bootstrap-hosted
+bootstrap-hosted: clone-hosted install-hosted copy-config install-crds install-resources kind-deploy-policy-framework kind-deploy-policy-controllers
+
+.PHONY: install-hosted
+install-hosted: 
+	@cd ./governance-policy-addon-controller && KIND_VERSION=latest HOSTED_MODE=true ./build/manage-clusters.sh
```
mprahl, line 360: Instead of hardcoding `KIND_VERSION=latest`, could you please use the value coming from `KIND_VERSION` in this Makefile?

.github/workflows/kind_host.yml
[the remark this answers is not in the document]
yiraeChristineKim: oh it looks working thanks! let me remove it

Makefile
```diff
@@ -349,3 +349,27 @@ integration-test:
 	else\
 		$(GINKGO) -v $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/integration;\
 	fi
+
+#hosted
+
+.PHONY: bootstrap-hosted
+bootstrap-hosted: clone-hosted install-hosted copy-config install-crds install-resources kind-deploy-policy-framework kind-deploy-policy-controllers
```
mprahl, line 356: `kind-deploy-policy-framework` and `kind-deploy-policy-controllers` deploy the controllers in normal mode. Refer back to our meeting recording to see the difference. In particular, the `ManagedClusterAddOn` object needs to have the `addon.open-cluster-management.io/hosting-cluster-name` annotation set. Additionally, it needs `installNamespace` to be set to something else.

Makefile
```diff
@@ -349,3 +349,27 @@ integration-test:
...
+.PHONY: 
+copy-config:
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
+
+.PHONY: clone-hosted
+clone-hosted:
+	-git clone --depth=1 https://github.com/open-cluster-management-io/governance-policy-addon-controller.git
```
mprahl, line 371: Could you please use the stolostron version since these tests aren't run in the upstream community yet?
```suggestion
	-git clone --depth=1 https://github.com/stolostron/governance-policy-addon-controller.git
```

mprahl, line 371: Could you please checkout the branch set to the `RELEASE_BRANCH` Makefile variable to use the consistent ACM version for the whole test?

Makefile
mprahl: Additionally, it seems that these tests are bypassing the governance-policy-addon-controller altogether, so you'll need to deploy the controllers first by installing the `governance-policy-addon-controller` on the hosted mode hub.

Also, remember that you need to create the secret with the Kubeconfig like I showed you in the recording for the controller on the management cluster to connect to the hosted/managed cluster.

mprahl:
@dhaiducek could you please also check this?

Makefile
yiraeChristineKim: If we use ManagedClusterAddOn like you showed before, we cannot pass the e2e test, because, for example, e2e test check policy copy in managed cluster. ManagedClusterAddOn way only creates something under policy and configurationpolicy

yiraeChristineKim: I suggest, if we want to test hosted mode in the way of "managedclusteraddon" as you showed me before, we need to write separate e2e tests for hosted mode

yiraeChristineKim:
Hi Team, after all comments are resolved, then I will squash the commits. Thanks!

go.mod
yiraeChristineKim: couldn't build hosted mode. when it creates a container, go mod install failed at k8s.io/client-go modules.

yiraeChristineKim: v 25 also error

Makefile
JustinKuli: > these tests are bypassing the governance-policy-addon-controller altogether, so you'll need to deploy the controllers first by installing the governance-policy-addon-controller on the hosted mode hub.

I wonder if that should be a separate PR/effort, possibly before we try to run these tests with a hosted cluster. 

> e2e test check policy copy in managed cluster

Right - currently, there are tests (and common test functions) that check that the replicated/propagated/child policies are copied to a cluster. For example: [here](https://github.com/stolostron/governance-policy-framework/blob/d6432a6757ea9f99bf750416778f359909658174/test/common/policy_utils.go#L107), [here](https://github.com/stolostron/governance-policy-framework/blob/main/test/common/policy_utils.go#L149), and [here](https://github.com/stolostron/governance-policy-framework/blob/main/test/configuration_policy_prune.go#L145). There are also tests that check for managed objects on a cluster, for example: [here](https://github.com/stolostron/governance-policy-framework/blob/main/test/configuration_policy_prune.go#L163), and [here](https://github.com/stolostron/governance-policy-framework/blob/main/test/e2e/configuration_policy_test.go#L117). Currently in both of those situations, they check the managed cluster. But, in hosted mode, the policies will be created on the "hosting cluster" (maybe usually the hub), while the managed objects are created on the "hosted cluster". My thought for this is that we have 3 kubeconfigs/clients, where currently we have 2:
1. A hub kubeconfig, like we currently have. Used for root policies, and verifying overall status.
2. A "hosting cluster" kubeconfig, which for our 'traditional' test runs would be the same as the managed kubeconfig. For new 'hosted mode' test runs, based on what I've heard here, this would actually be the hub cluster, but in the future we could potentially expand it to a separate cluster. This would be used for replicated policies and policy templates, and events.
3. A "target cluster" kubeconfig, which for our 'traditional' tests would be the same as the managed kubeconfig. This would be used to check that objects in a configuration policy were actually created, and for creating things like users and certificates to verify the functionality of iam/cert policies.

The test runs could always take all three of those via [configuration flags](https://github.com/stolostron/governance-policy-framework/blob/d6432a6757ea9f99bf750416778f359909658174/test/common/common.go#L48), and we could give the test runs the right values based on the type of the test. Then we would just need to update the current tests (see examples above) to use the correct clients where currently they always point to the managed cluster.

Makefile
```diff
@@ -349,3 +349,27 @@ integration-test:
...
+.PHONY: 
+copy-config:
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
+
+.PHONY: clone-hosted
+clone-hosted:
+	-git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git
```
dhaiducek, line 371: Optional: Similar to [targets in the addon-controller](https://github.com/stolostron/governance-policy-addon-controller/blob/9a2988e4a251ffff48cf23237014fce4694a4103/Makefile#L218-L221), this target could be: `$(ADDON_CONTROLLER)`, and then you can use that variable for the path throughout as well as not have to ignore when the clone fails (because it won't run at all since the directory already exists).

If you go this route, I'd probably also just make it a prerequisite for `install-hosted` so that that target can run successfully by itself.

```suggestion
ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
$(ADDON_CONTROLLER):
	git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git
```

Makefile
dhaiducek, line 356: Optional: I might prefer that this have the `kind-` prefix to match the other targets that deploy Kind.
```suggestion
kind-bootstrap-hosted: clone-hosted install-hosted copy-config install-crds install-resources kind-deploy-policy-framework kind-deploy-policy-controllers
```

Makefile
```diff
@@ -349,3 +349,27 @@ integration-test:
...
+.PHONY: install-hosted
+install-hosted: 
+	@cd ./governance-policy-addon-controller && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
+
+.PHONY: 
+copy-config:
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
```
dhaiducek, line 367: I think these targets should be combined since the goal is to deploy Kind clusters for the framework tests, I don't think there's a need to run them separately.
```suggestion
.PHONY: kind-install-hosted
kind-install-hosted: 
	@cd ./governance-policy-addon-controller && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
```

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,140 @@
...
+        export KUBECONFIG=${KUBECONFIG}
+        kubectl get deployment -A
+        kubectl get ns
+        echo "::endgroup::"
+    
+        echo "::group::test for patch step"
+        export MANAGED_CONFIG=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig
+        export HUB_CONFIG=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig
+        export HUB_CONFIG_INTERNAL=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal
+        echo "::endgroup::"
```
dhaiducek, line 86: I'm not sure how these fit in. Are these debug commands that will be removed?

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,140 @@
...
+
+        echo "::group::make setup hosted mode"
+        make bootstrap-hosted
+        echo "::endgroup::"
+        echo "::group::Saving kubeconfig paths for use in other steps"
+    
+        echo "MANAGED_KUBECONFIG=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig" >> $GITHUB_ENV
+        echo "HUB_KUBECONFIG=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig" >> $GITHUB_ENV
+        echo "HUB_INTERNAL_KUBECONFIG=$(pwd)/governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal" >> $GITHUB_ENV
```
dhaiducek, line 73: Since these are copied over to the root framework directory in the Makefile, it seems it might be clearer to use the kubeconfigs at the root of the repo rather than the ones from the addon controller?

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,140 @@
...
+        echo "::endgroup::"
+
+    - name: Patch Component Image
+      if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+      working-directory: component
+      env:
+        HUB_ONLY: ${{ inputs.hub_only_component }}
+        WATCH_NAMESPACE: managed
+      run: |
+        if [[ "${deployOnHub}" == "true" ]] || [[ "${HUB_ONLY}" == "true" ]]; then
```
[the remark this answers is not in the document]
yiraeChristineKim: good

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,140 @@
...
+      working-directory: framework
+      run: |
+        make e2e-debug-kind
+        tar -czvf e2e-debug.tar.gz test-output/debug
+    
+    - name: Upload Debug Artifacts
+      if: ${{ failure() }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: fw-kind-debug-${{ env.KIND }}-${{ env.DEPLOYONHUB }}
```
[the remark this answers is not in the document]
yiraeChristineKim: good

go.mod
```diff
@@ -10,13 +10,27 @@ require (
...
+	github.com/go-openapi/jsonreference v0.19.6 // indirect
+	github.com/go-openapi/swag v0.21.1 // indirect
+	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
+	github.com/google/gnostic v0.5.7-v3refs // indirect
+	github.com/josharian/intern v1.0.0 // indirect
+	github.com/mailru/easyjson v0.7.7 // indirect
+	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+)
+
 require (
```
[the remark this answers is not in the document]
yiraeChristineKim: I did `go mod tidy` um.... let me try.. I hope it will safe

Makefile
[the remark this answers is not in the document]
yiraeChristineKim: good catch

yiraeChristineKim: oh good I didn't know this Thanks!

Makefile
```diff
@@ -349,3 +349,24 @@ integration-test:
...
+
+$(ADDON_CONTROLLER):
+	git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git
+
+delete-hosted: 
+	@cd governance-policy-addon-controller && make kind-bootstrap-delete-clusters 
+	@rm kubeconfig_hub kubeconfig_managed kubeconfig_hub_internal kubeconfig_managed_internal
```
dhaiducek, line 372: Rather than adding the directory target to the bootstrap list, I'd prefer it be a prerequisite for the install-hosted target (that way, install-hosted can be run independently if desired):

```suggestion
ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller

.PHONY: kind-bootstrap-hosted
kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-deploy-policy-framework kind-deploy-policy-controllers

.PHONY: install-hosted
kind-install-hosted: $(ADDON_CONTROLLER)
	@cd $(ADDON_CONTROLLER) && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
	@cp $(ADDON_CONTROLLER)/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
	@cp $(ADDON_CONTROLLER)/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
	@cp $(ADDON_CONTROLLER)/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
	@cp $(ADDON_CONTROLLER)/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal

$(ADDON_CONTROLLER):
	git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git

delete-hosted: 
	@cd $(ADDON_CONTROLLER) && make kind-bootstrap-delete-clusters
	@rm kubeconfig_hub kubeconfig_managed kubeconfig_hub_internal kubeconfig_managed_internal

```

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,137 @@
...
+        echo "::endgroup::"
+
+    - name: Patch Component Image
+      if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+      working-directory: component
+      env:
+        HUB_ONLY: ${{ inputs.hub_only_component }}
+        WATCH_NAMESPACE: managed
+      run: |
+        if [[ "${HUB_ONLY}" == "true" ]]; then
```
mprahl, line 92: This if statement can be removed since we aren't doing "HUB_ONLY" in the hosted mode tests.

dhaiducek: @mprahl Isn't this referring to whether the component is hub only, like the propagator? https://github.com/stolostron/governance-policy-propagator/blob/main/.github/workflows/framework-kind.yaml

Makefile
mprahl: @JustinKuli thanks for the detailed suggestion!

@yiraeChristineKim could you please pursue this approach?

yiraeChristineKim: understand clear

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,140 @@
+name: Framework KinD hosted mode Workflow
```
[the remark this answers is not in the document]
yiraeChristineKim: um..... any suggestion?
I made a separate kind file for the future for that we can add more here. also to me, hosted mode is another stream of testing so...

Makefile
```diff
@@ -273,14 +275,24 @@ e2e-dependencies:
...
 K8SCLIENT ?= oc
 GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+CLUSTER_NAMESPACE_ON_HUB =
+TIMEOUT_SECONDS = 30
 .PHONY: e2e-test
 e2e-test:
 	@if [ -z "$(TEST_FILE)" ]; then\
-		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
+		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB) -timeout_seconds=$(TIMEOUT_SECONDS);\
```
yiraeChristineKim, line 284: In hosted mode, takes more time than normal more so I had to increase this number

Makefile
```diff
@@ -349,3 +361,51 @@ integration-test:
...
+kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-deploy-cert-manager setup-managedcluster
+
+.PHONY: kind-install-hosted
+kind-install-hosted: $(ADDON_CONTROLLER)
+	@cd ./governance-policy-addon-controller && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hub
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hub_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig ./kubeconfig_managed
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hosting
```
mprahl, line 379: Are the `kubeconfig_hosting` and `kubeconfig_hosting_internal` files used? I'm not seeing any reference to them. I think they are no longer used because of the new logic in `InitInterfaces`.

Makefile
```diff
@@ -27,6 +27,7 @@ export PATH := $(LOCAL_BIN):$(GOBIN):$(PATH)
 KIND_HUB_NAMESPACE ?= open-cluster-management
 KIND_MANAGED_NAMESPACE ?= open-cluster-management-agent-addon
 MANAGED_CLUSTER_NAME ?= managed
+HOSTING_CLUSTER_NAME ?= hosting
```
mprahl, line 30: This looks to be unused. Could you please remove it?

Makefile
```diff
@@ -349,3 +361,51 @@ integration-test:
 	else\
 		$(GINKGO) -v $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/integration;\
 	fi
+
+#hosted
+ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
+MANAGED_CLUSTER_NAME_IN_HOSTED = cluster2
```
mprahl, line 367: I think you can just use `MANAGED_CLUSTER_NAME` directly instead if you stop overriding `MANAGED_CLUSTER_NAME`.

Makefile
```diff
@@ -349,3 +361,51 @@ integration-test:
...
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hosting
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hosting_internal
+	@echo installing policy-propagator on hub
+	-kubectl create ns $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/governance-policy-propagator/$(RELEASE_BRANCH)/deploy/operator.yaml -n $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/iam-policy-controller/$(RELEASE_BRANCH)/deploy/crds/policy.open-cluster-management.io_iampolicies.yaml --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/cert-policy-controller/$(RELEASE_BRANCH)/deploy/crds/policy.open-cluster-management.io_certificatepolicies.yaml --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+
+tmp:
```
mprahl, line 387: Can you remove this Make target?

Makefile
```diff
@@ -349,3 +361,51 @@ integration-test:
...
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl2.kubeconfig-internal ./kubeconfig_managed_internal
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig ./kubeconfig_hosting
+	@cp ./governance-policy-addon-controller/policy-addon-ctrl1.kubeconfig-internal ./kubeconfig_hosting_internal
+	@echo installing policy-propagator on hub
+	-kubectl create ns $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/governance-policy-propagator/$(RELEASE_BRANCH)/deploy/operator.yaml -n $(KIND_HUB_NAMESPACE) --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/iam-policy-controller/$(RELEASE_BRANCH)/deploy/crds/policy.open-cluster-management.io_iampolicies.yaml --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
+	kubectl apply -f https://raw.githubusercontent.com/stolostron/cert-policy-controller/$(RELEASE_BRANCH)/deploy/crds/policy.open-cluster-management.io_certificatepolicies.yaml --kubeconfig=$(PWD)/kubeconfig_$(HUB_CLUSTER_NAME)
```
mprahl, line 385: The addon controller should install these CRDs. Could you please delete these lines?

Makefile
```diff
@@ -349,3 +361,51 @@ integration-test:
...
+$(ADDON_CONTROLLER):
+	git clone --depth=1 -b $(RELEASE_BRANCH) https://github.com/stolostron/governance-policy-addon-controller.git
+
+.PHONY: setup-managedcluster
+setup-managedcluster: 
+	-kubectl create ns $(MANAGED_CLUSTER_NAME_IN_HOSTED)-hosted --kubeconfig=$(PWD)/kubeconfig_hub
+	-kubectl -n $(MANAGED_CLUSTER_NAME_IN_HOSTED)-hosted create secret generic config-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=$(PWD)/kubeconfig_hub
+	-kubectl -n $(MANAGED_CLUSTER_NAME_IN_HOSTED)-hosted create secret generic cert-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=$(PWD)/kubeconfig_hub
+	-kubectl -n $(MANAGED_CLUSTER_NAME_IN_HOSTED)-hosted create secret generic iam-policy-controller-managed-kubeconfig --from-file=kubeconfig=$(PWD)/kubeconfig_managed_internal --kubeconfig=$(PWD)/kubeconfig_hub
+	-kubectl apply -f test/resources/hosted_mode/managed-cluster-addon.yaml --kubeconfig=$(PWD)/kubeconfig_hub -n $(MANAGED_CLUSTER_NAME_IN_HOSTED)
```
mprahl, line 401: Could you please follow the other convention of `kubeconfig_$(HUB_CLUSTER_NAME)` instead of hardcoding `kubeconfig_hub`?

test/common/policy_utils.go
```diff
@@ -213,21 +214,23 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
 }
 
 func DoHistoryUpdatedTest(policyName string, messages ...string) {
-	By("Getting policy history")
+	By("Getting policy history, check latest message")
 	Eventually(func(g Gomega) {
 		history, _, err := GetHistoryMessages(policyName, 0)
-		g.Expect(err).Should(BeNil())
+		g.Expect(err).ShouldNot(HaveOccurred())
```
mprahl, line 220: Could you please explain why you made this change?

test/common/policy_utils.go
```diff
@@ -213,21 +214,23 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
...
 		historyMsgs := []string{}
 		for _, h := range history {
 			historyItem, _ := h.(map[string]interface{})
 			m, _, _ := unstructured.NestedString(historyItem, "message")
 			historyMsgs = append(historyMsgs, m)
 		}
-		By("Check history length is same")
-		g.Expect(len(history)).Should(Equal(lenMessage))
+		By("Check history length is same or has more than expected")
+		g.Expect(len(history) >= lenMessage).Should(BeTrue())
```
mprahl, line 229: Could you please describe the situation where this change is needed?

test/common/policy_utils.go
```diff
@@ -213,21 +214,23 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
...
 		By("Check history message same")
-		g.Expect(strings.Join(historyMsgs, "")).Should(Equal(strings.Join(messages, "")))
+		histroymss := historyMsgs[0:lenMessage]
+		fmt.Println(histroymss)
```
mprahl, line 232:
```suggestion
```

Makefile
yiraeChristineKim: I tested several times with default timeout, iam and cert policy test had fallen to fail... but set to 60secs working well. I guess hosted mode takes more time than normal one

mprahl: If you follow my comment about setting `MANAGED_CLUSTER_NAME` to `cluster2`, then we shouldn't need this new variable of `MANAGED_CLUSTER_NAME_IN_HOSTED`.

Makefile
```diff
@@ -229,14 +229,23 @@ install-resources:
...
 
 K8SCLIENT ?= oc
+GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+CLUSTER_NAMESPACE_ON_HUB =
 .PHONY: e2e-test
 e2e-test:
 	@if [ -z "$(TEST_FILE)" ]; then\
-		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
+		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB);\
```
mprahl, line 238: Could you use a new variable called `MANAGED_CLUSTER_NAMESPACE` and have it default to `MANAGED_CLUSTER_NAME`? I find it confusing that `MANAGED_CLUSTER_NAME` is set to `cluster2-hosted` because its actual name is still `cluster2`, it's just its installed namespace on the hosting cluster is `cluster2-hosted`.

Makefile
mprahl: I think you should still be able to use `MANAGED_CLUSTER_NAME` instead of this new variable.

Makefile
mprahl: This comment still applies. Please remove these CRD installations. The addon controller should already handle this.

test/configuration_policy_prune.go
```diff
@@ -29,41 +31,84 @@ func ConfigPruneBehavior(labels ...string) bool {
...
-			GinkgoWriter.Printf("cleanPolicy OcManaged policy event output: %v\n", outManaged)
-			Expect(err).To(BeNil())
-			outManaged, err = OcManaged(
-				"delete", "events", "-n", ClusterNamespace,
-				"--field-selector=involvedObject.name="+policyName,
-				"--ignore-not-found",
+			By("Removing any finalizers from the configmap")
+
+			_, _ = OcManaged("patch", "configmap", pruneConfigMapName, "-n", "default",
+				"--type=json", "-p", `[{"op":"remove", "path":"/metadata/finalizers"}]`)
```
mprahl, line 45: Why are these changes needed to the prune test?

yiraeChristineKim: in hosted mode, deleting events interrupts testings I guess it has event latency or something

test/configuration_policy_prune.go
```diff
@@ -366,32 +433,17 @@ func ConfigPruneBehavior(labels ...string) bool {
...
 			"default",
 			!cmShouldBeDeleted,
-			DefaultTimeoutSeconds,
+			120,
 		)
 	}
 
 	Describe("GRC: [P1][Sev1][policy-grc] "+
 		"Test configuration policy pruning", Ordered, Label(labels...), func() {
-		cleanConfigMap := func() {
```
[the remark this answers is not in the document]
mprahl: We don't support running the tests in parallel. Although it's a good idea to do so, it'd be out of scope of this PR. Could you please revert the changes to the prune test that aren't related to using the hosting cluster client or using the correct namespace on the hub?

test/configuration_policy_prune.go
mprahl: What do you mean by interrupts testing? The tests should only run serially.

.github/workflows/e2e.yml
```diff
@@ -11,6 +11,10 @@ on:
       - release-2.[7-9]
 
 jobs:
+  kind-host:
```
mprahl, line 14:
```suggestion
  kind-hosted-mode:
```

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,130 @@
...
```
mprahl, line 1:
```suggestion
name: Framework KinD Hosted Mode Workflow
```

Makefile
```diff
@@ -229,14 +229,23 @@ install-resources:
...
 	else\
 		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast --focus-file=$(TEST_FILE) test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
```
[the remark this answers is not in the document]
mprahl: This still applies. Could you please ensure that `-cluster_namespace=$(MANAGED_CLUSTER_NAMESPACE) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB)` are passed here also?

test/common/policy_utils.go
mprahl: What are the additional messages? This sounds like it could be a bug.

test/policy_ordering.go
```diff
@@ -78,19 +96,21 @@ func PolicyOrdering(labels ...string) bool {
 				DoRootComplianceTest(policyWithDepName, policiesv1.NonCompliant)
 			})
 			It("Should become pending again when the initial policy is deleted", func() {
+				cleanEvent()
```
mprahl, line 99: @JustinKuli could you please provide your input on this?

I don't think we want to include this here.

test/policy_ordering.go
```diff
@@ -78,19 +96,21 @@ func PolicyOrdering(labels ...string) bool {
...
-		Describe("Ordering via an extraDependency on a ConfigurationPolicy", func() {
+		Describe("Ordering via an extraDependency on a ConfigurationPolicy", Ordered, func() {
 			BeforeAll(func() {
 				By("Creating the initial policy to use as a dependency")
 				DoCreatePolicyTest(initialPolicyYaml, GvrConfigurationPolicy)
 				DoRootComplianceTest(initialPolicyName, policiesv1.NonCompliant)
 
 				By("Creating the policy that depends on the initial policy")
 				DoCreatePolicyTest(policyWithExtraDepYaml)
+				cleanEvent()
```
mprahl, line 113: Same as the previous comment. I don't think we want this here since you are artificially bringing the policy to pending. I'll defer to @JustinKuli on this though.

test/policy_ordering.go
```diff
@@ -101,19 +121,21 @@ func PolicyOrdering(labels ...string) bool {
...
-		Describe("IgnorePending should allow policies to be compliant when one template is pending", func() {
+		Describe("IgnorePending should allow policies to be compliant when one template is pending", Ordered, func() {
 			BeforeAll(func() {
 				By("Creating the initial policy to use as a dependency")
 				DoCreatePolicyTest(initialPolicyYaml, GvrConfigurationPolicy)
 				DoRootComplianceTest(initialPolicyName, policiesv1.NonCompliant)
 
 				By("Creating the policy that depends on the initial policy")
 				DoCreatePolicyTest(ignorePendingYaml)
+				cleanEvent()
```
mprahl, line 138: Same as above

Makefile
```diff
@@ -233,14 +233,25 @@ install-resources:
...
+GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+TIMEOUT_SECONDS=30
+MANAGED_CLUSTER_NAMESPACE ?= cluster2-hosted
+
 .PHONY: e2e-test
 e2e-test:
 	@if [ -z "$(TEST_FILE)" ]; then\
-		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAME) -k8s_client=$(K8SCLIENT) ;\
+		$(GINKGO) -v --no-color $(TEST_ARGS) --fail-fast test/e2e -- -cluster_namespace=$(MANAGED_CLUSTER_NAMESPACE) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(CLUSTER_NAMESPACE_ON_HUB) -timeout_seconds=$(TIMEOUT_SECONDS);\
```
mprahl, line 244: What fails with the reduced timeout in hosted mode? I think it might be because the controllers are installed asynchronously by the addon controller in hosted mode whereas in normal mode, the Makefile waits for the pods to be up.

I think if you add to the Makefile to wait for the controllers to be running on the hosting cluster, you might not need this additional timeout increase.

test/common/gvr.go
```diff
@@ -13,6 +13,10 @@ var (
 		Version:  "v1",
 		Resource: "namespaces",
 	}
+	GvrSecret = schema.GroupVersionResource{
```
mprahl, line 16: Could you please delete this? It seems to be unused.

Makefile
yiraeChristineKim: what is suggestion? it is not failed during the installation. it is failed during cert and iam e2e test

yiraeChristineKim: let me take a screen shot for u

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,130 @@
...
+        echo "HUB_INTERNAL_KUBECONFIG=$(pwd)/kubeconfig_hub_internal" >> $GITHUB_ENV
+        echo "::endgroup::"
+    
+        echo "::group::test for patch step"
+        export MANAGED_CONFIG=$(pwd)/kubeconfig_managed
+        export HUB_CONFIG=$(pwd)/kubeconfig_hub
+        export HUB_CONFIG_INTERNAL=$(pwd)/kubeconfig_hub_internal
+        echo "::endgroup::"
+
+    - name: Patch Component Image
```
mprahl, line 85: I don't think this will work in hosted mode. I think you likely want to run `make build-images` and then run something like `kind load docker-image $(REGISTRY)/$(IMG):$(TAG) --name $(KIND_NAME)` on the Hub cluster so that the new image is used. You may have to manually trigger a "rollout" of the controller here as well.

Makefile
```diff
@@ -233,14 +233,25 @@ install-resources:
 e2e-dependencies:
 
 K8SCLIENT ?= oc
+GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+TIMEOUT_SECONDS=30
+MANAGED_CLUSTER_NAMESPACE ?= cluster2-hosted
```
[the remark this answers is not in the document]
mprahl: @yiraeChristineKim this comment still applies.

test/resources/hosted_mode/managed-cluster-addon.yaml
```diff
@@ -0,0 +1,39 @@
...
+    addon.open-cluster-management.io/hosting-cluster-name: cluster1
+spec:
+  installNamespace: cluster2-hosted
+---
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ManagedClusterAddOn
+metadata:
+  name: config-policy-controller
+  annotations:
+    addon.open-cluster-management.io/hosting-cluster-name: cluster1
```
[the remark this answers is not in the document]
yiraeChristineKim: understand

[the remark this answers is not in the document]
yiraeChristineKim: oh what about  `addon.open-cluster-management.io/values: '{"args": {"frequency": 10}}' should I replace?
`

Makefile
```diff
@@ -233,14 +233,24 @@ install-resources:
 e2e-dependencies:
 
 K8SCLIENT ?= oc
+GINKGO = $(LOCAL_BIN)/ginkgo
+IS_HOSTED ?= false
+MANAGED_CLUSTER_NAMESPACE ?= managed
```
mprahl, line 238: Do you need to set a default of `CLUSTER_NAMESPACE_ON_HUB` such as?
```suggestion
MANAGED_CLUSTER_NAMESPACE ?= managed
CLUSTER_NAMESPACE_ON_HUB ?= $(MANAGED_CLUSTER_NAMESPACE)
```

.github/workflows/kind_host.yml
```diff
@@ -0,0 +1,138 @@
...
+        # `repository` is inferred as the "caller" repository 
+        path: framework
+        # `ref` is inferred as the new commit
+
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: framework/go.mod
+
+    - name: Bootstrap the KinD Cluster for hosted mode
```
mprahl, line 60:
```suggestion
    - name: Bootstrap the KinD clusters for hosted mode
```

test/common/policy_utils.go
mprahl: @yiraeChristineKim are these changes still needed?

mprahl (commented):
Just a few minor comments but this is looking awesome!

Makefile
yiraeChristineKim: common.go I set ```
func InitInterfaces(hubConfig, managedConfig string, isHosted bool) {
	if isHosted {
		ClientHosting = NewKubeClient("", hubConfig, "")
		ClientHostingDynamic = NewKubeClientDynamic("", hubConfig, "")
	} else {
		ClientHosting = NewKubeClient("", managedConfig, "")
		ClientHostingDynamic = NewKubeClientDynamic("", managedConfig, "")
	}

	if ClusterNamespaceOnHub == "" {
		ClusterNamespaceOnHub = ClusterNamespace
	}

	...
```
actually you set this on integration_suite_test.go e2e_suite_test.go but I moved to there because it is duplicated

File: Makefile (hunk 41 above)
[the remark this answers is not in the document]
Comment (yiraeChristineKim) [reply Makefile]: I changed to `MANAGED_CLUSTER_NAMESPACE ?=managed`

File: .github/workflows/kind_host.yml (hunk 44 above)
Comment (yiraeChristineKim) [reply .github/workflows/kind_host.yml:60]: gotcha

File: .github/workflows/kind_host.yml [hunk 45]
```diff
@@ -0,0 +1,138 @@
...
+        echo "Use $VERSION_TAG"
+
+        echo " KIND_NAME=policy-addon-ctrl1 and the hub config as the default kubeconfig"
+
+        echo "Using KIND_NAME=policy-addon-ctrl1 and the hub config as the default kubeconfig"
+        export KIND_NAME=policy-addon-ctrl1
+        export KUBECONFIG=${HUB_KUBECONFIG}
+
+        echo "::group::make build-images"
+        make build-images
```
mprahl, line 104:
```suggestion
        TAG=$(VERSION_TAG) make build-images
```

test/e2e/configuration_policy_test.go
```diff
@@ -20,24 +20,30 @@ import (
...
+	_, err = common.OcHosting(
+		"delete", "events", "-n", common.ClusterNamespace,
+		"--field-selector=involvedObject.name="+rolePolicyName,
 		"--ignore-not-found",
 	)
 	ExpectWithOffset(1, err).To(BeNil())
-	_, err = common.OcManaged(
-		"delete", "events", "-n", "managed",
+	_, err = common.OcHub(
+		"delete", "events", "-n", common.UserNamespace,
```
mprahl, line 36:
```suggestion
		"delete", "events", "-n", common.ClusterNamespaceOnHub,
```

mprahl, line 36: Did you mean this?
```suggestion
	_, err = common.OcHosting(
		"delete", "events", "-n", common.ClusterNamespace,
```

yiraeChristineKim: I was trying to remove this event `policy-test       role-policy-musthave               inform               NonCompliant       51s`

mprahl: There shouldn't be any events published to `UserNamespace` though. It should only be in the managed cluster namespace.

yiraeChristineKim: let me test without that let you know

mprahl (approved):
Nice!

PR #575, opened 2023-03-10
CI check script updates

dhaiducek:
- Drop `2.4`
- Drop `grc-ui`/`grc-ui-api`
- Add policy generator and policy collection

### DIFF
--- a/build/periodic.sh
+++ b/build/periodic.sh
@@ -4,7 +4,7 @@
 
 COMPONENT_ORG=stolostron
 DEFAULT_BRANCH=${DEFAULT_BRANCH:-"main"}
-CHECK_RELEASES="2.4 2.5 2.6 2.7 2.8"
+CHECK_RELEASES="2.5 2.6 2.7 2.8"
 # This list can include all postsubmit jobs for all repos--if a job doesn't exist it's filtered to empty and skipped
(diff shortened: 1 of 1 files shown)


build/periodic.sh
```diff
@@ -22,11 +22,11 @@ cloneRepos() {
...
 		# Manually append deprecated repos
 		REPOS="${REPOS}
-			stolostron/grc-ui
-			stolostron/grc-ui-api
 			stolostron/governance-policy-spec-sync
 			stolostron/governance-policy-status-sync
-			stolostron/governance-policy-template-sync"
+			stolostron/governance-policy-template-sync
+			stolostron/policy-collection
+			stolostron/policy-generator-plugin"
```
mprahl, line 29: Should we also add `governance-policy-framework-addon`?

dhaiducek: No. It's already in `repo.txt`: https://github.com/stolostron/policy-grc-squad/blob/35f8a8646b751ce8a8039b938a4e3dbee3aad051/main-branch-sync/repo.txt#L6

dhaiducek: These are the repos that either don't have an image build or have dropped from our current version.

PR #578, opened 2023-03-17
Bump test timeouts

dhaiducek:
E2E framework tests and canaries were failing--this should resolve the failures. I had some concerns the new framework-addon code was the cause, but it's not across the board and some of these tests already had long timeouts, so if I have time I might look into it, but I no longer feel that investigating the framework-addon code is quite as high a priority.

Also bumps the propagator version to pick up the EventuallyWithOffset test util changes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,5 +57,5 @@ require (
 replace (
 	golang.org/x/text => golang.org/x/text v0.3.8 // CVE-2022-32149
 	k8s.io/client-go => k8s.io/client-go v0.23.9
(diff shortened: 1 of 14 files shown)


gparvin (approved):

dhaiducek:
The E2E tests have failed here on every run with some variation of the status not being returned/updated on different tests. I might have to go ahead and investigate this after all.

gparvin (approved):

dhaiducek:
Closing for now. Looks like the failures were resolved by:
- https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/67

PR #581, opened 2023-03-21
Check the managed cluster for the multiline tests

dhaiducek:
When run in the canaries, this policy is getting deployed to the Managed cluster, but all the checks were against the hub cluster.

Also does some refactoring to loop over the hub and managed clusters.

ref: https://github.com/stolostron/backlog/issues/27232

### DIFF
--- a/test/integration/policy_multiline_templatization_test.go
+++ b/test/integration/policy_multiline_templatization_test.go
@@ -5,12 +5,14 @@ package integration
 
 import (
 	"context"
+	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


dhaiducek:
CI here is passing. This PR is ready for review.

test/integration/policy_multiline_templatization_test.go
```diff
@@ -68,43 +67,31 @@ var _ = Describe(
...
-					if err != nil {
-						return ""
-					}
-
-					return configMap.Data["extraData"]
-				},
-				defaultTimeoutSeconds*2,
-				1,
-			).Should(Equal("exists!"))
+		It("The ConfigMaps on the managed cluster should be patched with the correct data", func() {
```
mprahl, line 70: Should this also verify the Hub results as well?

dhaiducek: I don't think so because there's no guarantee that the hub is self-managed.

dhaiducek: It's an interesting point. I'll add in the check, but it does make me wonder whether we ever test a non-self-managed hub with a managed cluster. I guess not...

mprahl (approved):

PR #582, opened 2023-03-21
Pass a pointer for OCPUser

dhaiducek:
If subscription admin setup (specifically, logging in) fails, cleanup doesn't have the populated OCPUser to work with. Using a pointer populates the values so that cleanup can succeed in that case.

### DIFF
--- a/test/common/gitops_utils.go
+++ b/test/common/gitops_utils.go
@@ -18,20 +18,18 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// GitOpsUserSetup configures a new user to use for the GitOps tests.
-// It returns the OCPUser instance, which contains a path to the created kubeconfig file.
-func GitOpsUserSetup() OCPUser {
+// GitOpsUserSetup configures a new user to use for the GitOps tests. It updates the provided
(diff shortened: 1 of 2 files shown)


dhaiducek:
CI here is passing. This PR is ready for review.

test/common/gitops_utils.go
```diff
@@ -20,18 +20,16 @@ import (
 
 // GitOpsUserSetup configures a new user to use for the GitOps tests.
 // It returns the OCPUser instance, which contains a path to the created kubeconfig file.
-func GitOpsUserSetup() OCPUser {
+func GitOpsUserSetup(ocpUser *OCPUser) {
```
JustinKuli, line 23: Since the instance isn't returned, this comment should be updated.

JustinKuli (changes requested):
Seems good overall! Just one small thing.

test/common/gitops_utils.go
dhaiducek: Good point! Done!

mprahl (approved):

JustinKuli (approved):

PR #591, opened 2023-03-29
Wait for the compliance messages in the configuration policy tests

mprahl:
Prior to this, the test would check the compliance messages at the end of the test, however, this could run into race conditions. For example, if a policy becomes compliant due to updating an object, an additional status message can be added if the Config Policy controller manages to perform another evaluation before the test moves on.

This check adds the status message checks at each step so that the test waits for all expected status messages to appear before moving on.

This should unblock the CI failure here:
https://github.com/stolostron/governance-policy-framework-addon/pull/53

There seemed to be race conditions encountered where consecutive tests
    that used the same policy, would sometimes have status events from the
    previous test get added even if the events were deleted. This is likely
    due to a delay in the Status Sync's cache of events. To fix this, I
    rearranged the tests to be two Describe blocks, one for the inform tests
    and another for the enforce tests. This now makes it so that consecutive
    tests don't use the same policy.

### DIFF
--- a/test/common/policy_utils.go
+++ b/test/common/policy_utils.go
@@ -215,6 +215,12 @@ func GetLatestStatusMessage(policyName string, templateIdx int) func() string {
 
 func DoHistoryUpdatedTest(policyName string, messages ...string) {
 	By("Getting policy history")
+
+	// There is a limit of 10 messages in the Policy status, so if more are passed in, just truncate it.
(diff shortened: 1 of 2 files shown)


yiraeChristineKim:
This is good but why only changed this part, why don't you change all?

yiraeChristineKim (approved):

PR #592, opened 2023-04-04
Polish ConfigurationPolicy pruning test

dhaiducek:
It seems that if the ConfigurationPolicy hadn't been successfully updated with a status (even after it had successfully created the ConfigMap), it was vulnerable to deletion, so this adds a check for its status to be up-to-date.

Additionally, there was no timeout for deleting the ConfigMap with a finalizer, so the command would continue to wait indefinitely. This adds a timeout as well as a patch to remove the finalizer.


test/configuration_policy_prune.go
```diff
@@ -28,12 +28,28 @@ func ConfigPruneBehavior(labels ...string) bool {
 			GinkgoWriter.Printf("cleanPolicy OcHub output: %v\n", outHub)
 			Expect(err).To(BeNil())
 
+			Eventually(func(g Gomega) {
+				outManaged, err := OcManaged(
+					"get", "configurationpolicies", "-A",
+				)
+				GinkgoWriter.Printf("cleanPolicy OcManaged configurationpolicy output: %v\n", outManaged)
+				Expect(outManaged).To(BeEmpty())
+				Expect(err).To(BeNil())
```
mprahl, line 37:
```suggestion
				g.Expect(outManaged).To(BeEmpty())
				g.Expect(err).To(BeNil())
```

PR #594, opened 2023-04-04
Test ManagedClusterLabels in hub templates

JustinKuli:
One test is added to the BVT, one is SVT-only. I was planning on having an additional test to check that if a label is queried that doesn't exist, then the policy would be non-compliant (I thought that such an error would appear as a template-error), but the Go stdlib `template` inserts `<no value>` in that situation. That might not be ideal, but it's not something to fix at this moment.

Refs:
 - https://issues.redhat.com/browse/ACM-4177

### DIFF
--- a/test/integration/hubtemplates_managedclusterlabels_test.go
+++ b/test/integration/hubtemplates_managedclusterlabels_test.go
@@ -1,0 +1,112 @@
+// Copyright (c) 2023 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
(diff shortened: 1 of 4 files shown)


mprahl (approved):

JustinKuli:
This test really hates me

mprahl:
The prune test should be more stable after this fix:
https://github.com/open-cluster-management-io/config-policy-controller/pull/118

I need to file a follow-up Jira to make the code around handling status update failures more robust in general when pruneObjectBehavior is used.

PR #604, opened 2023-04-20
Fix CI workflow

dhaiducek:
- Fix JSON formatting
- Use separate step to post results (the previous code wasn't called because the CI commands exited with an error code)


dhaiducek:
CI is passing--this is ready for review.

gparvin (approved):

PR #613, opened 2023-05-11
Merge hosted mode tests into the usual workflow

JustinKuli:
Most steps were already the same. This should just add one new case to the matrix.

Refs:
 - https://issues.redhat.com/browse/ACM-5211
 - https://issues.redhat.com/browse/ACM-5212


JustinKuli:
Companion Prow Config PR: https://github.com/openshift/release/pull/39279

dhaiducek (dismissed):
/hold if you want to address the nits. Otherwise, this is a great consolidation! Thanks for working this out!

Makefile
```diff
@@ -325,6 +330,8 @@ ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
 
 .PHONY: kind-bootstrap-hosted
 kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-deploy-cert-manager setup-managedcluster
+	kubectl delete pod -l name=governance-policy-propagator -A --kubeconfig=./kubeconfig_hub
+	kubectl delete pod -l app=governance-policy-addon-controller -A --kubeconfig=./kubeconfig_hub
```
yiraeChristineKim, line 334: why delete addon controller?

.github/workflows/kind.yml
```diff
@@ -88,46 +99,69 @@ jobs:
...
         HUB_ONLY: ${{ inputs.hub_only_component }}
-        WATCH_NAMESPACE: managed
+        HOSTED: ${{ matrix.hosted }}
       run: |
+        if [[ "$RELEASE_BRANCH" != "main" ]]; then
+          export TAG=latest-$(echo $RELEASE_BRANCH | cut -d'-' -f 2):
+        else
+          export TAG=latest
+        fi
+
```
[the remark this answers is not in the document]
JustinKuli: I think it was @yiraeChristineKim that put this in the hosted workflow, so credit goes to her!

Makefile
JustinKuli: I just needed to restart it. In my first test runs, it kept having problems when running in the workflow, but not when I was running it locally. When I added this, the tests started passing, I'm not 100% sure why 😆

.github/workflows/kind.yml
```diff
@@ -88,46 +99,69 @@ jobs:
       env:
         deployOnHub: ${{ matrix.deployOnHub }}
         KIND_VERSION: ${{ matrix.kind }}
+        HOSTED: ${{ matrix.hosted }}
```
dhaiducek, line 102: Nit: I don't think this env is used outside of the workflow, so there's not a real benefit to defining an env here rather than using `${{ matrix.hosted }}` directly? (Same with the E2E step)

JustinKuli: I prefer to not use github expressions inside the `run`. When I do that, I worry that I haven't written it correctly, and I'll get a weird syntax error either in GH or the shell. Keeping the expressions in the `env` block means the `run` block is a valid, standalone script.

.github/workflows/kind.yml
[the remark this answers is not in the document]
yiraeChristineKim: Matt caught this issue lol I just added   ` echo $RELEASE_BRANCH | cut -d'-' -f 2`

Makefile
yiraeChristineKim: ok!! but so weird I haven't seen that 
oh maybe  `PLACEMENT_TAG=latest-amd64 make kind-bootstrap-hosted ` Are you mac?

Makefile
```diff
@@ -325,6 +330,9 @@ ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
 
 .PHONY: kind-bootstrap-hosted
 kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-deploy-cert-manager setup-managedcluster
+	@echo "Restarting propagator and addon-controller"
```
dhaiducek, line 333: I've started doing `# Restarting propagator and addon-controller` instead of `@echo`. Doesn't really matter, but I kind of like the free marker at the beginning and not needing to ignore the command.

dhaiducek (approved):

dhaiducek:
@JustinKuli You're welcome to unhold this from my perspective if you're ready. I've approved the `release` PR also.

Makefile
JustinKuli: Apparently my editor doesn't know if that `#` really means a comment in a makefile. So I was worried I would break some syntax somewhere 😆

PR #619, opened 2023-05-31
Remove tagging the image

dhaiducek:
Since the image is being built and tagged locally only in component repos, I'm not sure setting a tag is necessary.


yiraeChristineKim (approved):
let's give it a try

PR #625, opened 2023-06-22
Restart the governance-policy-framework addon after Gatekeeper is uninstalled

mprahl:
This is to disable the Gatekeeper status controller. This would happen automatically, but it can take a while since it relies on the health endpoint.

Relates:
https://issues.redhat.com/browse/ACM-5965


JustinKuli (approved):
/override "Framework KinD / Tests (latest, false, hosted)"

yiraeChristineKim (approved):

PR #626, opened 2023-06-26
Remove the Kyverno policy test

mprahl:
The Kyverno Helm chart has gotten too large to be supported by Application Subscription, so the policy that installs Kyverno can no longer be used.

See the following comment:
https://github.com/stolostron/backlog/issues/27329#issuecomment-1602619144


dhaiducek (approved):

mprahl:
/override "Framework KinD / Tests (latest, false, hosted)"

PR #628, opened 2023-07-04
Fix hosted mode error on component framework test

yiraeChristineKim:
Description:  a reusable GitHub workflow for running the Kind E2E tests in hosted mode was added. We need to now trigger these tests from the component repos as part of the PR CI in releases 2.7+.

Modify the governance-policy-framework release-2.7, release-2.8, and main branches to consolidate this as a new job in the existing reusable workflow file.

Ref: https://issues.redhat.com/browse/ACM-5211


.github/workflows/kind.yml
```diff
@@ -30,170 +30,177 @@ jobs:
...
-          - kind: 'latest'
-            deployOnHub: 'false'
-            hosted: 'hosted'
+          - kind: "latest"
+            deployOnHub: "false"
+            hosted: "hosted"
 
     steps:
-    - name: Checkout Component Repository
-      # Checkout a specific component when called from another repository
```
yiraeChristineKim, line 52: These chages are lints by VScode

.github/workflows/kind.yml
```diff
@@ -30,170 +30,177 @@ jobs:
...
+          echo "::endgroup::"
+
+      - name: Patch Component Image
+        if: ${{ github.event.repository.name != 'governance-policy-framework' }}
+        working-directory: component
+        env:
+          deployOnHub: ${{ matrix.deployOnHub }}
+          HUB_ONLY: ${{ inputs.hub_only_component }}
+          HOSTED: ${{ matrix.hosted }}
+          REPO: ${{ github.event.repository.name }}
```
yiraeChristineKim, line 137: Changes start!

.github/workflows/kind.yml
```diff
@@ -134,16 +134,23 @@ jobs:
...
         export HUB_CONFIG=${HUB_KUBECONFIG}
         export HUB_CONFIG_INTERNAL=${HUB_INTERNAL_KUBECONFIG}
 
         if [[ "${HOSTED}" == "hosted" ]]; then
           export WATCH_NAMESPACE=cluster2-hosted
-          export KIND_NAMESPACE=cluster2-hosted
           export KIND_NAME=policy-addon-ctrl1
           export KUBECONFIG=${HUB_KUBECONFIG}
+          export KIND_NAMESPACE=cluster2-hosted
+          if [[ "${REPO}" == "governance-policy-propagator" ]]; then
```
mprahl, line 148: Instead of this workflow choosing the namespace, could the repo calling this workflow explicitly override the namespace? In other words, something in the `governance-policy-propagator` GitHub workflow file explicitly sets it.

This would be cleaner in my opinion.

yiraeChristineKim:
```
      if [ ! "${KIND_NAMESPACE}" ]; then
            export KIND_NAMESPACE=cluster2-hosted
       fi
```
Then I will add `KIND_NAMESPACE `env at propagator repo/framework-kind.yaml
How about this?

dhaiducek: We should be able to use `HUB_ONLY` to conditionally set the namespace to `open-cluster-management` for hub components. Would that cover it?

yiraeChristineKim: I think it is better set env in each repo's makefile,  config, iam, cert vs propagator, addon framework-kind.yaml(which is in each repo) don't have information about hosted mode so

yiraeChristineKim: I added like this in each repo.
```
	if [ "$(HOSTED)" -eq "hosted" ]; then\
		export KIND_NAMESPACE=open-cluster-management ;\
	else 
		$(MAKE) kind-deploy-controller;\
	fi;
```

dhaiducek: I'm not so sure. The "cluster2-hosted" name is defined this this repo, so it seems that it makes more sense to set it here also rather than in each repo?

mprahl:
@yiraeChristineKim could you please explain why this fixes the issue? I'm a little behind on this PR discussion.  :smile:

.github/workflows/kind.yml
yiraeChristineKim: In propagator controller KIND NAMESPCE should be` open-cluster-management` and addon controller should be `governance-policy-addon-controller-system `

yiraeChristineKim:
@mprahl  @dhaiducek  So each controller KIND NAMESPACE should be different. KIND_NAMESPCE  Cert, Iam, Config controller are in` cluster2-hosted ` These are made by managedclusteraddon. However, governance-policy-propagator is in `open-cluster-management` namespace. Moreover,  policy-addon-ctrl-controller-manager is in` policy-addon-ctrl-controller-manager `namespace. I made our hosted mode e2e test environment based on addoncontroller.

yiraeChristineKim:
> > @mprahl @dhaiducek So each controller KIND NAMESPACE should be different. KIND_NAMESPCE Cert, Iam, Config controller are in`cluster2-hosted` These are made by managedclusteraddon. However, governance-policy-propagator is in `open-cluster-management` namespace. Moreover, policy-addon-ctrl-controller-manager is in`policy-addon-ctrl-controller-manager`namespace. I made our hosted mode e2e test environment based on addoncontroller.
> 
> Thanks for the explanation. I think we should simplify it so that the propagator/addon-controller (hub components signaled by the `HUB_ONLY` flag) are deployed to `open-cluster-management` (that should already be configurable for the addon-controller?) and the other components default to `cluster2-hosted` so that namespaces can be set here in a central location for the hosted mode test.

@dhaiducek  Do you have an idea how to simplify?

dhaiducek:
> @dhaiducek Do you have an idea how to simplify?

I've dug a bit more into it. It looks like the addon-controller could use the same treatment as the config-policy-controller where it's using a `CONTROLLER_NAMESPACE` rather than a `KIND_NAMESPACE` variable. After that's updated, I'm thinking invoke the addon-controller hosted deployment with `KIND_NAMESPACE=$(KIND_HUB_NAMESPACE)` on this line:
https://github.com/stolostron/governance-policy-framework/blob/d77d9dc9b33665ca9d895ffb843cf9d088106b78/Makefile#L339

Then, in the workflow, we can maybe do something like:
```shell
        if [[ "${HOSTED}" == "hosted" ]]; then
          if [[ "${HUB_ONLY}" == "true" ]]; then
            export KIND_NAMESPACE=open-cluster-management
          else
            export KIND_NAMESPACE=cluster2-hosted
          fi
          export WATCH_NAMESPACE=cluster2-hosted
          export KIND_NAME=policy-addon-ctrl1
          export KUBECONFIG=${HUB_KUBECONFIG}
        elif [[ "${deployOnHub}" == "true" ]] || [[ "${HUB_ONLY}" == "true" ]]; then
          export WATCH_NAMESPACE=managed
          export KIND_NAME=hub
          export KUBECONFIG=${HUB_KUBECONFIG}
        else
          export WATCH_NAMESPACE=managed
          export KIND_NAME=managed
          export KUBECONFIG=${MANAGED_KUBECONFIG}
        fi
```

Would that make sense?

yiraeChristineKim:
> KIND_HUB_NAMESPACE

Overall I get it but KIND_HUB_NAMESPACE, where is this from?

dhaiducek:
> @dhaiducek Overall I get it but KIND_HUB_NAMESPACE, where is this from? Oh!! `Addon-ctrl `need`governance-policy-addon-controller-system` namespace `propagator` need `open-cluster-management `namespace

It's from the framework Makefile:
https://github.com/stolostron/governance-policy-framework/blob/d77d9dc9b33665ca9d895ffb843cf9d088106b78/Makefile#L31

yiraeChristineKim:
@dhaiducek         
```
   if [[ "${HUB_ONLY}" == "true" ]]; then
            export KIND_NAMESPACE=open-cluster-management
   else
```
This part is only for `propagator controller . What about addon controller?

dhaiducek:
> This part is only for `propagator controller . What about addon controller?

I'm suggesting that we adjust the Makefile in the addon-controller repo so that this works there also.

yiraeChristineKim:
> > This part is only for `propagator controller . What about addon controller?
> 
> I'm suggesting that we adjust the Makefile in the addon-controller repo so that this works there also.

So Can you correct if I am wrong? so my understanding is adding this `export KIND_NAMESPACE=governance-policy-addon-controller-system` in the Addon controller?

dhaiducek:
> So Can you correct if I am wrong? so my understanding is adding this `export KIND_NAMESPACE=governance-policy-addon-controller-system` in the Addon controller? In other words, set KIND_NAMESPACE env only in Addon controller

Sort of, but in the Makefile so that it can be overridden from the framework like the other repos:
```make
CONTROLLER_NAMESPACE ?= governance-policy-addon-controller-system
```
to
```make
KIND_NAMESPACE ?= governance-policy-addon-controller-system
```
https://github.com/stolostron/governance-policy-addon-controller/blob/52c545245504f9032cba2bff1c5f325498de8e27/Makefile#L257

yiraeChristineKim:
@Dale so You want to set `KIND_NAMESPACE ?= governance-policy-addon-controller-system` and `KIND_NAMESPACE = governance-policy-addon-controller-system`  in addon-cotroller makefile?

dhaiducek:
> Now clear @dhaiducek Thanks again. In propagator-controller, I will change to` KIND_NAMESPACE ?= governance-policy-addon-controller-system`

You mean "In the **_addon_**-controller", right? 🙂

dhaiducek (approved):
/hold for consensus

@mprahl I'm good with this--are you fine with this solution?

Asking because in https://github.com/stolostron/governance-policy-framework/pull/628#discussion_r1253039254, it was originally mentioned to set the namespace in each repo, but I'm not sure that makes sense since I think as much as possible hosted mode should be set up by whatever is calling the controller (in this case this workflow)?

mprahl (approved):

mprahl:
> /hold for consensus
> 
> @mprahl I'm good with this--are you fine with this solution?
> 
> Asking because in [#628 (comment)](https://github.com/stolostron/governance-policy-framework/pull/628#discussion_r1253039254), it was originally mentioned to set the namespace in each repo, but I'm not sure that makes sense since I think as much as possible hosted mode should be set up by whatever is calling the controller (in this case this workflow)?

Thanks for pulling me in! This is better than what I suggested and it accounts for my intention, which was for input from the workflow caller (i.e. component repo) that determines the namespace.

PR #634, opened 2023-07-11
Add oauth deployment checks

dhaiducek:
Attempts to fix the flaky GitOps user login by monitoring the oauth deployment for readiness

Also uses the `.go/` directory for the addon-controller so that it's hidden and not at the root of the repo when it exists.

ref: https://github.com/stolostron/backlog/issues/27338

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -326,7 +326,7 @@ integration-test:
 	fi
 
 #hosted
-ADDON_CONTROLLER = $(PWD)/governance-policy-addon-controller
+ADDON_CONTROLLER = $(PWD)/.go/governance-policy-addon-controller
 
 .PHONY: kind-bootstrap-hosted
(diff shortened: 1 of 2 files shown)


yiraeChristineKim:
What is the reason move to under `./go`

dhaiducek:
> What is the reason move to under `./go`

If the Make target fails, the repo folder won't accidentally be committed or monitored because it's in a directory that's hidden and in the `.gitignore`:
https://github.com/stolostron/governance-policy-framework/blob/d77d9dc9b33665ca9d895ffb843cf9d088106b78/.gitignore#L10

There's precedence for this in the addon-controller: 
https://github.com/open-cluster-management-io/governance-policy-addon-controller/blob/8edfa35f191c134fe16a7d8c56a62bacc5baf5d0/Makefile#L218

JustinKuli (approved):

PR #644, opened 2023-07-14
Adjust E2E default configurations

dhaiducek:
- Shorten debug log output (The dump, now that it is fixed, was causing Travis to run out of logging space)
- Remove `--fail-fast` as a default Ginkgo argument 
- Adjust E2E regex to include all releases - (It used to be that we didn't want it to run on 2.6, but now we do want it to run on 2.9+)
- Make `TEST_FILE` usage cleaner

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -4,11 +4,11 @@ on:
   pull_request:
     branches:
       - main
-      - release-2.[7-9]
+      - release-[0-9]+.[0-9]+
   push:
     branches:
       - main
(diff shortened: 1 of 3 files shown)


dhaiducek:
The Travis job for reference:
```
The job exceeded the maximum log length, and has been terminated.
```
https://app.travis-ci.com/github/stolostron/governance-policy-framework/jobs/606048378#L19997

50 lines might be too conservative, but I'm not sure how much we rely on on the dump anyway.

JustinKuli (approved):
LGTM!

PR #654, opened 2023-08-08
Add script to detect Go modules

dhaiducek:
Sample output:
```
$ ./detect-packages.sh goproxy
Go module query: goproxy
* Checking stolostron/cert-policy-controller
CGO_ENABLED=1 go build -o ./build/_output/bin/cert-policy-controller ./main.go
=== stolostron/cert-policy-controller 2.6: goproxy not found
CGO_ENABLED=1 go build -o ./build/_output/bin/cert-policy-controller ./main.go
=== stolostron/cert-policy-controller 2.7: goproxy not found
CGO_ENABLED=1 go build -o ./build/_output/bin/cert-policy-controller ./main.go
=== stolostron/cert-policy-controller 2.8: goproxy not found
CGO_ENABLED=1 go build -o ./build/_output/bin/cert-policy-controller ./main.go
=== stolostron/cert-policy-controller 2.9: goproxy not found
. . .
```

~It lacks a summary currently, but~ it definitely speeds up searching for modules in our binaries. 🙂

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -14,7 +14,8 @@
 
 # Utility scripts
 
-| Folder                                   | Description                                                                                          |
-| ---------------------------------------- | ---------------------------------------------------------------------------------------------------- |
-| [branch-create/](./branch-create/)       | Scripts for cutting new release branches and configurations                                          |
-| [main-branch-sync/](./main-branch-sync/) | Scripts for maintaining cross-repo `main` branches and fast-forwarding to the current release branch |
+| Folder                                    | Description                                                                                          |
+| ----------------------------------------- | ---------------------------------------------------------------------------------------------------- |
+| [branch-create/](./branch-create/)        | Scripts for cutting new release branches and configurations                                          |
+| [main-branch-sync/](./main-branch-sync/)  | Scripts for maintaining cross-repo `main` branches and fast-forwarding to the current release branch |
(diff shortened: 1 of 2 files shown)


mprahl:
@dhaiducek could you use `go list -m all` instead to avoid building the binary?

dhaiducek:
> @dhaiducek could you use `go list -m all` instead to avoid building the binary?

Yes, that ought to work. This was a way to rule out false positives if a module didn't actually wind up in the binary, but `go list` would definitely be faster...

dhaiducek:
> > @dhaiducek could you use `go list -m all` instead to avoid building the binary?
> 
> Yes, that ought to work. This was a way to rule out false positives if a module didn't actually wind up in the binary, but `go list` would definitely be faster...

Just tried it with `goproxy`. Building the binary and using `go version` didn't find it in any version. `go list` finds it in every repo in every version.

mprahl:
> There ought to be a way around that, though. I'll try to see how `go list` needs to be fiddled with to get only the main modules...

This gnarly command may do it?
https://stackoverflow.com/a/64390523

dhaiducek:
There was a `jq` version in that post that I found to be much cleaner. 🙂

dhaiducek:
/unhold

I pulled out the Go template from the one-liner to make it more readable. It's actually not as scary as it seemed 😄

PR #660, opened 2023-09-29
Adjust encryption test to use new variable

dhaiducek:
I have no idea whether this will do anything, but it might... 🤷‍♂️ 

ref: https://issues.redhat.com/browse/ACM-4873


mprahl (approved):
I doubt this will fix it but it doesn't hurt.

dhaiducek:
> I doubt this will fix it but it doesn't hurt.

Agreed, especially since it's a pointer. My other thought was to extend the timeout, but I don't think that's it either since the propagator says the token was rotated and we're fetching it directly from the hub. It's a mysterious flake.

PR #662, opened 2023-10-04
Fix report file name

dhaiducek:
- Fixes an E2E report filename
- Also reverts this PR since it's not necessary:
  - #660

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -197,7 +197,7 @@ jobs:
     - name: Upload Test Reports
       uses: actions/upload-artifact@v3
       with:
-        name: fw-kind-report-${{ matrix.kind }}-${{ matrix.deployOnHub }}${{ matrix.hosted == 'hosted' && '-hosted' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/kind.yml
```diff
@@ -197,7 +197,7 @@ jobs:
     - name: Upload Test Reports
       uses: actions/upload-artifact@v3
       with:
-        name: fw-kind-report-${{ matrix.kind }}-${{ matrix.deployOnHub }}${{ matrix.hosted == 'hosted' && '-hosted' }}
+        name: fw-kind-report-${{ matrix.kind }}-${{ matrix.deployOnHub }}${{ matrix.hosted == 'hosted' && '-hosted' || '' }}
```
yiraeChristineKim, line 200: Can we use like this  matrix.hosted == 'hosted' ? '-hosted' : ''  (2. Ternary Operator or condition operator)

JustinKuli: This is the closest I think we can get to a real ternary operator in github actions syntax :/ https://docs.github.com/en/actions/learn-github-actions/expressions#example

yiraeChristineKim: Thanks!

yiraeChristineKim (approved):

PR #664, opened 2023-10-12
Update the controller names from the propagator

JustinKuli:
A change in the propagator split up the `policy-propagator` controller into a 2 controllers: `root-policy-spec`, and `replicated-policy`. So the old metrics are no longer reported, and these tests should check for the new ones.

Refs:
 - https://github.com/open-cluster-management-io/governance-policy-propagator/pull/130


dhaiducek (approved):
Thanks, @JustinKuli!

JustinKuli:
I think this one has now failed twice:
```
Summarizing 1 Failure:
  [FAIL] GRC: [P1][Sev1][policy-grc] Test the ACM Hardening generated PolicySet in an App subscription [It] Validates the propagated policies [policy-collection, stable]
  /go/src/github.com/stolostron/governance-policy-framework/test/integration/policy_generator_acm_hardening_test.go:124
```

JustinKuli:
Something seems to be happening with a PolicySet. The test that is failing is because a policy in a set is not being distributed to `local-cluster`. The binding/placement that should make that happen are attached to the `acm-hardening` PolicySet.

Based on the logs from the propagator, when it tries to look up that PolicySet (from a reference in a PlacementBinding, I think), it gets a "not found". And earlier, the policyset-controller logs a similar error, on its reconciles for that PolicySet:
```
2023-10-12T15:09:20.882Z	info	policy-set	policyset/policyset_controller.go:49	Reconciling policy sets...	{"Request.Namespace": "policies", "Request.Name": "acm-hardening"}
2023-10-12T15:09:20.882Z	info	policy-set	policyset/policyset_controller.go:59	Policy set not found, so it may have been deleted.	{"Request.Namespace": "policies", "Request.Name": "acm-hardening"}
```

dhaiducek:
Oh my goodness. It finally worked. 🎉 🤯 Good work! Do you want to keep the debug in there or take it out before merging it?

dhaiducek (dismissed):
/hold (in case the debug logs should be removed)

JustinKuli:
I feel like I'm tempting fate by removing the debugs logs and trying again, but here it goes!

PR #667, opened 2023-10-24
Fix flaky complicance operator test

yiraeChristineKim:
The compliance operator test sometimes fails because the scan doesn't complete within the timeout. The check for DONE should be removed to speed up the test and reduce flakiness.

Ref: https://issues.redhat.com/browse/ACM-8238

### DIFF
--- a/test/integration/policy_comp_operator_test.go
+++ b/test/integration/policy_comp_operator_test.go
@@ -132,22 +132,7 @@ func complianceScanTest(scanPolicyName string, scanPolicyURL string, scanName st
 				)
 
 				return compliancesuite.Object["status"].(map[string]interface{})["phase"]
-			}, common.MaxTimeoutSeconds, 1).Should(Equal("AGGREGATING"))
-		})
-		It("ComplianceSuite "+scanName+" scan results should be DONE", func() {
-			By("Checking if ComplianceSuite " + scanName + " scan status.phase is DONE")
-			Eventually(func() interface{} {
-				compliancesuite := utils.GetWithTimeout(
-					clientManagedDynamic,
-					common.GvrComplianceSuite,
-					scanName,
-					"openshift-compliance",
-					true,
-					defaultTimeoutSeconds,
-				)
-
-				return compliancesuite.Object["status"].(map[string]interface{})["phase"]
-			}, common.MaxTimeoutSeconds, 1).Should(Equal("DONE"))
+			}, common.MaxTimeoutSeconds, 1).Should(Or(Equal("AGGREGATING"), Equal("RUNNING")))
 		})
 	})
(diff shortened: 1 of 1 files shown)


test/integration/policy_comp_operator_test.go
```diff
@@ -226,7 +226,7 @@ func complianceScanTest(scanPolicyName string, scanPolicyURL string, scanName st
 	})
 }
 
-var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] "+
+var _ = FDescribe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] "+
```
mprahl, line 229:
```suggestion
var _ = Describe("RHACM4K-2222 GRC: [P1][Sev1][policy-grc] "+
```

test/integration/policy_comp_operator_test.go
```diff
@@ -132,7 +132,7 @@ func complianceScanTest(scanPolicyName string, scanPolicyURL string, scanName st
 				)
 
 				return compliancesuite.Object["status"].(map[string]interface{})["phase"]
-			}, common.MaxTimeoutSeconds, 1).Should(Equal("AGGREGATING"))
+			}, common.MaxTimeoutSeconds, 1).Should(Or(Equal("AGGREGATING"), Equal("RUNNING")))
 		})
 		It("ComplianceSuite "+scanName+" scan results should be DONE", func() {
```
mprahl, line 137: I think we also want to remove this `It`.

yiraeChristineKim: you mean.. 
```
		It("ComplianceSuite "+scanName+" scan results should be AGGREGATING", func() {
			By("Checking if ComplianceSuite " + scanName + " scan status.phase is AGGREGATING")
			Eventually(func() interface{} {
				compliancesuite := utils.GetWithTimeout(
					clientManagedDynamic,
					common.GvrComplianceSuite,
					scanName,
					"openshift-compliance",
					true,
					defaultTimeoutSeconds,
				)

				return compliancesuite.Object["status"].(map[string]interface{})["phase"]
			}, common.MaxTimeoutSeconds, 1).Should(Or(Equal("AGGREGATING"), Equal("RUNNING")))
		})
``` ?

dhaiducek: @mprahl is referring to the next `It()` checking for `DONE`, since if it's still running, it also won't be done for a while and we should just move on.

yiraeChristineKim: oh after this is clean up so `DONE` is meaningless ok!! Let me change

mprahl (approved):

PR #675, opened 2023-12-01
Fix debug log collection for hosted mode framework E2E tests

mprahl:


.github/workflows/kind.yml
```diff
@@ -184,7 +184,16 @@ jobs:
...
+        if [[ "${HOSTED}" == "hosted" ]]; then
+        KIND_MANAGED_NAMESPACE=cluster2-hosted \
+          MANAGED_CLUSTER_NAME=hub \
+          KIND_COMPONENT_SELECTOR=app \
+          KIND_COMPONENTS='config-policy-controller cert-policy-controller iam-policy-controller governance-policy-framework' \
+          make e2e-debug-kind
+        else
+          make e2e-debug-kind
+        fi
+
```
dhaiducek, line 196: I'm not 100% sure on this, but it looks like this would work? (Without needing to update the Makefile?)
```suggestion
        if [[ "${HOSTED}" == "hosted" ]]; then
          KIND_MANAGED_NAMESPACE=cluster2-hosted \
          MANAGED_CLUSTER_NAME=hub \
          make e2e-debug-acm
        else
          make e2e-debug-kind
        fi

```

mprahl: @dhaiducek the problem is that in hosted mode, it's deployed using the addon controller and I didn't see a label of `name`, so I had to override that to `app`.

Then for whatever reason, the label value is `app: governance-policy-framework` but when not in the addon controller, it's `name: governance-policy-framework-addon`. That's why I had to override those variables.

mprahl: Ah, missed that the make target changed in your diff.

dhaiducek: Yeah, the change in target is why I'm not 100% on it, but using the addon controller would probably mean it's closer to ACM, and it looks like it'd work as we'd like...

dhaiducek (dismissed):
/hold In case you want to set up the debug to run in the action always() to see whether it works before releasing it into the wild. 🙂

Thanks for fixing this!

dhaiducek (approved):

PR #679, opened 2023-12-07
Output CI results to Slack

dhaiducek:
Before I had concerns about the volume of information, but seeing the integration test output combined with the reduced output here it's time to go ahead and output the information to Slack directly.


JustinKuli (approved):
LGTM

PR #682, opened 2023-12-12
Fix E2E PlacementRef in Subscriptions

dhaiducek:
I forgot to update the `placementRef` in the Subscription 🤦‍♂️ 

Followup to:
- #681


gparvin (approved):
Thanks for fixing the tests

PR #683, opened 2023-12-12
Remove subscription allow list

dhaiducek:
It's unnecessary and was blocking Placement from being deployed.

### DIFF
--- a/test/resources/policy_generator/subscription.yaml
+++ b/test/resources/policy_generator/subscription.yaml
@@ -43,16 +43,6 @@ metadata:
     app: grc-e2e-policy-generator
   name: grc-e2e-policy-generator-subscription
 spec:
-  allow:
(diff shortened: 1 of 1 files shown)


JustinKuli (commented):
Did it used to be necessary?

dhaiducek:
> Did it used to be necessary?

There was a time early on where AppSub was considering an allow list and we preemptively added it, but the behavior is if there isn't a list, it allows everything:
```
                // If allow list is empty, all resources are allowed for deploy
                if len(allowlist) == 0 {
                        return true
                }
```
https://github.com/open-cluster-management-io/multicloud-operators-subscription/blob/dcce8c6074b95a5747fed8cab1daf20772c8605c/pkg/utils/subscription.go#L733-L736

JustinKuli (approved):

PR #687, opened 2023-12-14
Fix hosted test workflow

JustinKuli:
Commit 1: Based on the example at https://docs.github.com/en/actions/learn-github-actions/variables#using-the-vars-context-to-access-configuration-variable-values. I don't know why double quotes doesn't work, but it's worth a try.

Commit 2: Apply the ClusterManagementAddOn resources required for the addon-controller, see https://github.com/open-cluster-management-io/governance-policy-addon-controller/commit/b3d6cf14aa0b620bf47af4f0cef0f97c0e4b6f62

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -71,7 +71,7 @@ jobs:
     name: Fast Forward
     needs: [ e2e, integration ]
     runs-on: ubuntu-latest
-    if: ${{ vars.FAST_FORWARD == "true" }}
+    if: ${{ vars.FAST_FORWARD == 'true' }}
     steps:
     - name: Checkout Policy Framework
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


dhaiducek (approved):
Works for me, though has the double quotes even been tested yet? I haven't seen a passing run since the PR merged...

JustinKuli:
@dhaiducek right, I should've provided the context that the workflow has not been run since the previous change was merged, it just has this error:
```
The workflow is not valid. .github/workflows/integration.yml (Line: 74, Col: 9): Unexpected symbol: '"true"'. Located at position 22 within expression: vars.FAST_FORWARD == "true"
```

dhaiducek:
Oh, curious. Thanks! I hadn't noticed that one (would have been nice if it had flagged it in the PR!). I wonder if it's because it's using Go under the hood and Go uses double quotes, so they're not compatible or something. 🤔 

> @dhaiducek right, I should've provided the context that the workflow has not been run since the previous change was merged, it just has this error

gparvin (approved):

PR #725, opened 2024-02-15
Add integration tests: Helm in the PolicyGenerator

dhaiducek:
Confirmed that `kustomize` internally uses `helm template`:
https://github.com/kubernetes-sigs/kustomize/blob/91ccf00ac8f787551eb4c5baad486b088ef3737b/api/types/helmchartargs.go#L149

Which doesn't call the Kubernetes API, so security here should be fine, but we have a `lookup` in the chart just to make sure:
```shell
$ helm template --help
Render chart templates locally and display the output.

Any values that would normally be looked up or retrieved in-cluster will be
faked locally. Additionally, none of the server-side testing of chart validity
(e.g. whether an API is supported) is done.
```

ref: https://issues.redhat.com/browse/ACM-9456

### DIFF
--- a/test/integration/policy_generator_helm_test.go
+++ b/test/integration/policy_generator_helm_test.go
@@ -1,0 +1,131 @@
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the Policy Generator "+
+	"with a Helm chart", Ordered, Label("SVT"), func() {
+	const policyName = "e2e-grc-helm-policy-app"
+	const namespace = "grc-e2e-helm-policy-generator"
+
+	It("Sets up the application subscription", func(ctx SpecContext) {
+		By("Creating the application subscription")
+		_, err := common.OcUser(
+			gitopsUser,
+			"apply",
+			"-f",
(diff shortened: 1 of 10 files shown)


dhaiducek:
I've upgraded the E2E cluster, so this PR should be good to merge once CI passes.

test/integration/policy_generator_helm_test.go
```diff
@@ -0,0 +1,133 @@
...
+			g.Expect(err).ShouldNot(HaveOccurred(), "The subscription status should be parseable.")
+			g.Expect(found).Should(BeTrue(), "The subscription status should have a phase.")
+			g.Expect(appSubPhase).Should(Equal("Propagated"), "The subscription should propagate successfully.")
+		},
+			defaultTimeoutSeconds,
+			1,
+		).Should(Succeed())
+	})
+
+	It("Validates the propagated policies", func() {
```
yiraeChristineKim, line 49:
```suggestion
	It("Validates the propagated policies", func(ctx SpecContext) {
```

Optional

test/integration/policy_generator_helm_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// Perform some basic validation on the generated policy.
+		By("Checking that the root policy was created")
+		policyRsrc := clientHubDynamic.Resource(common.GvrPolicy)
+		var policy *unstructured.Unstructured
+		Eventually(
+			func() error {
+				var err error
+				policy, err = policyRsrc.Namespace(namespace).Get(
+					context.TODO(), policyName, metav1.GetOptions{},
```
yiraeChristineKim, line 58: Optional: ctx SpecContext
```suggestion
					ctx, policyName, metav1.GetOptions{},
```

yiraeChristineKim (dismissed):

test/integration/policy_generator_helm_test.go
dhaiducek: Oh wow. I didn't even realize that was a thing! Thanks!

yiraeChristineKim (approved):

PR #727, opened 2024-02-19
Identify the metrics svc/route by name

dhaiducek:
The label is no longer unique with the compliance history API coming on board, and I don't think there's much risk of the name being changed.

ref: https://github.com/stolostron/backlog/issues/27486


JustinKuli (approved):
LGTM

PR #729, opened 2024-02-21
Fix ci status check

dhaiducek:
The source is definitely some unescaped quotation marks in the JSON, so we'll see whether escaping them works...


JustinKuli (approved):
`¯\_(ツ)_/¯`

PR #736, opened 2024-02-26
Add tests for installing a specific operator via OperatorPolicy

zyjjay:
ref: https://issues.redhat.com/browse/ACM-9894

### DIFF
--- a/test/integration/policy_install_operator_test.go
+++ b/test/integration/policy_install_operator_test.go
@@ -1,0 +1,401 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/test/utils"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+var _ = Describe("GRC: [P1][Sev1][policy-grc] Test install Operator",
+	Ordered, Label("BVT"), func() {
+		const (
+			testNS              = "grcqeoptest-ns"
+			policyNoGroupYAML   = "../resources/policy_install_operator/operator_policy_no_group.yaml"
+			policyWithGroupYAML = "../resources/policy_install_operator/operator_policy_with_group.yaml"
+			policyNamePrefix    = "test-op"
+			noGroupSuffix       = "-43544"
+			withGroupSuffix     = "-43545"
+			csvName             = "quay-operator.v3.8.15"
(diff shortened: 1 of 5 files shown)


test/integration/policy_install_operator_test.go
```diff
@@ -0,0 +1,389 @@
...
+				By("Checking the policy on managed cluster in the namespace " + clusterNamespace)
+				managedPolicy := utils.GetWithTimeout(
+					clientManagedDynamic,
+					common.GvrPolicy,
+					userNamespace+"."+policyNamePrefix+noGroupSuffix,
+					clusterNamespace,
+					true,
+					defaultTimeoutSeconds,
+				)
+				Expect(managedPolicy).NotTo(BeNil())
```
JustinKuli, line 108: I think you could use `common.DoCreatePolicyTest` for this, although you would need to adjust the name of the file, policy, and template (IIRC they all need to match).

test/integration/policy_install_operator_test.go
```diff
@@ -0,0 +1,389 @@
...
+			It("Should verify CSV details", func() {
+				csv := utils.GetWithTimeout(
+					clientManagedDynamic,
+					common.GvrClusterServiceVersion,
+					csvName,
+					testNS+noGroupSuffix,
+					true,
+					defaultTimeoutSeconds*4,
+				)
+				Expect(csv).NotTo(BeNil())
```
JustinKuli, line 210: Are there any details in the CSV we want to check? Maybe this whole thing could be in an Eventually waiting for a certain status?

JustinKuli (dismissed):
/hold

Looks good to me overall! I think it could be nice to verify some additional details, even though that isn't specified in the test scenario. But we don't necessarily have to add anything to this.

test/integration/policy_install_operator_test.go
```diff
@@ -0,0 +1,389 @@
...
+					defaultTimeoutSeconds,
+				)
+				Expect(managedPolicy).NotTo(BeNil())
+			})
+
+			It("operator-policy"+noGroupSuffix+" should be NonCompliant", func() {
+				By("Checking if the status of the root policy is NonCompliant")
+				Eventually(
+					common.GetComplianceState(policyNamePrefix+noGroupSuffix),
+					defaultTimeoutSeconds*2,
```
JustinKuli, line 115: Lots of these tests use defaultTimeoutSeconds*2. Is that necessary? It would be nice to use the default for most cases, although I understand that some of the tests here might need to wait longer for OLM to create something.

test/integration/policy_install_operator_test.go
JustinKuli: Oh, right, the names here probably match exactly what is in the test scenario. That's a good reason to keep it the way it is.

JustinKuli (dismissed):
/unhold

LGTM

JustinKuli:
/hold

let's see if we can get 2 test passes in a row here to try and ensure some stability.

JustinKuli:
/unhold

I see two green runs, thanks!

PR #737, opened 2024-02-27
Add BVT tests for operator policy errors

JeffeyL:
ref: https://issues.redhat.com/browse/ACM-9895

### DIFF
--- a/test/integration/operator_policy_error_test.go
+++ b/test/integration/operator_policy_error_test.go
@@ -1,0 +1,126 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+var _ = Describe("GRC: [P1][Sev1][policy-grc] Test operatorpolicy errors",
+	Ordered, Label("BVT"), func() {
+		const (
+			policyPath            = "../resources/operator_policy/test-op-err-initial.yaml"
+			correctPolicyPath     = "../resources/operator_policy/test-op-err-correct.yaml"
+			policyName            = "test-op-err-initial"
+			operatorPolicyName    = "test-op-err-initial"
+			subscriptionNamespace = "grcqeoptest-ns-43568"
+		)
+
+		BeforeAll(func() {
+			By("Create policy " + policyName + " which contains " + operatorPolicyName)
+			common.DoCreatePolicyTest(policyPath, common.GvrOperatorPolicy)
+		})
+
+		AfterAll(func() {
+			_, err := common.OcHub(
+				"delete", "-f", policyPath,
+				"-n", userNamespace, "--ignore-not-found=true",
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = common.OcManaged(
+				"delete", "ns", subscriptionNamespace, "--ignore-not-found=true",
+			)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should be NonCompliant due to invalid namespace", func() {
+			By("Checking if the status of the operator policy is NonCompliant")
+			Eventually(
+				common.GetComplianceState(policyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(Equal(policiesv1.NonCompliant))
+			Eventually(
+				common.GetOpPolicyCompMsg(operatorPolicyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(ContainSubstring("the operator namespace ('grcqeoptest-notcreated') does not exist"))
+		})
+
+		It("Namespace "+subscriptionNamespace+" should be created", func() {
+			_, err := common.OcManaged(
+				"create", "namespace", subscriptionNamespace,
+			)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should be patched with correct policy", func() {
+			_, err := common.OcHub(
+				"apply", "-f", correctPolicyPath, "-n", userNamespace,
+			)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should be patched with incorrect source", func() {
+			_, err := common.OcHub(
+				"patch", "policies.policy.open-cluster-management.io", policyName,
+				"-n", userNamespace, "--type=json", "-p", `[{
+				"op":"replace", 
+				"path":"/spec/policy-templates/0/objectDefinition/spec/subscription/source",
+				"value":"redhat-operators-invalid"
+			}]`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should be NonCompliant due to incorrect source", func() {
+			By("Checking if the status of the operator policy is NonCompliant")
+			Eventually(
+				common.GetComplianceState(policyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(Equal(policiesv1.NonCompliant))
+			Eventually(
+				common.GetOpPolicyCompMsg(operatorPolicyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(ContainSubstring("CatalogSource 'redhat-operators-invalid' was not found"))
+		})
+
+		It(operatorPolicyName+" should be patched with correct policy", func() {
+			_, err := common.OcHub(
+				"apply", "-f", correctPolicyPath, "-n", userNamespace,
+			)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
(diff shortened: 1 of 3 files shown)


test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,193 @@
...
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should have namespace removed", func() {
+			_, err := common.OcHub(
+				"patch", "policies.policy.open-cluster-management.io", policyName,
+				"-n", userNamespace, "--type=json", "-p", `[{
+				"op":"remove", 
+				"path":"/spec/policy-templates/0/objectDefinition/spec/subscription/namespace",
+				"value":"Invalid"
```
JustinKuli, line 175: I think this line is effectively ignored by the patch

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,193 @@
...
+			}]`)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It(operatorPolicyName+" should be NonCompliant", func() {
+			By("Checking if the status of the operator policy is NonCompliant")
+			Eventually(
+				common.GetComplianceState(policyName),
+				defaultTimeoutSeconds*2,
```
JustinKuli, line 148: Lots of these tests use `defaultTimeoutSeconds*2`. Is that necessary? It would be nice to use the default for most cases, although I understand that some of the tests here might need to wait longer for OLM to create something.

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,193 @@
...
+			By("Checking the policy on managed cluster in ns " + clusterNamespace)
+			managedPolicy := utils.GetWithTimeout(
+				clientManagedDynamic,
+				common.GvrPolicy,
+				userNamespace+"."+policyName,
+				clusterNamespace,
+				true,
+				defaultTimeoutSeconds,
+			)
+			Expect(managedPolicy).NotTo(BeNil())
```
JustinKuli, line 62: I think you could use `common.DoCreatePolicyTest` for this, although you would need to adjust the name of the file, policy, and template (IIRC they all need to match).

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,193 @@
...
+		It(operatorPolicyName+" should be patched with incorrect installPlanApproval", func() {
+			_, err := common.OcHub(
+				"patch", "policies.policy.open-cluster-management.io", policyName,
+				"-n", userNamespace, "--type=json", "-p", `[{
+				"op":"replace", 
+				"path":"/spec/policy-templates/0/objectDefinition/spec/subscription/installPlanApproval",
+				"value":"Invalid"
+			}]`)
+			Expect(err).ToNot(HaveOccurred())
+		})
```
JustinKuli, line 142: Between checking the status for something specific, it looks like this suite patches the policy twice: once to undo a previous change, then it applies a new detail to check. It seems like it could be slightly nicer to replace that first patch with an "apply" back to the known state. That way, if one of the patch commands is adjusted in the future, it's corresponding "undo" doesn't need to be updated - the "apply" will always work and set the policy to something known.

JustinKuli (commented):
Seems good overall. I have some thoughts, mostly for maintainability.

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+var _ = FDescribe("GRC: [P1][Sev1][policy-grc] Test operatorpolicy errors",
```
JustinKuli, line 14:
```suggestion
var _ = Describe("GRC: [P1][Sev1][policy-grc] Test operatorpolicy errors",
```

JustinKuli (changes requested):
Other than the FDescribe, I think this looks good. Thanks for the updates

JustinKuli:
/retest ci/prow/test-e2e-grc-framework

I want to make sure this is pretty stable. I think it is.

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,157 @@
...
+				defaultTimeoutSeconds,
+				1,
+			).Should(ContainSubstring("CatalogSource 'redhat-operators-invalid' was not found"))
+		})
+
+		It(operatorPolicyName+" should be patched with correct policy", func() {
+			_, err := common.OcHub(
+				"apply", "-f", correctPolicyPath, "-n", userNamespace,
+			)
+			Expect(err).ToNot(HaveOccurred())
```
yiraeChristineKim, line 99: To me it is good to add verifying the `correctPolicyPath` is correct one 

`common.GetComplianceState = Compliant` like this

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,157 @@
...
+			Eventually(
+				common.GetOpPolicyCompMsg(operatorPolicyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(ContainSubstring("the policy spec.subscription.installPlanApproval ('Invalid') is invalid"))
+		})
+
+		It(operatorPolicyName+" should be patched with correct policy", func() {
+			_, err := common.OcHub(
+				"apply", "-f", correctPolicyPath, "-n", userNamespace,
```
yiraeChristineKim, line 129: same here add verifying

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,157 @@
...
+				defaultTimeoutSeconds,
+				1,
+			).Should(Equal(policiesv1.NonCompliant))
+			Eventually(
+				common.GetOpPolicyCompMsg(operatorPolicyName),
+				defaultTimeoutSeconds,
+				1,
+			).Should(ContainSubstring("namespace is required in spec.subscription"))
+		})
+	})
```
yiraeChristineKim, line 157: Can we add error for `startingCSV` not existing version?

yiraeChristineKim:
Wow overall very clean and easy to understand!

test/resources/operator_policy/test-op-err-correct.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: test-op-err-initial-plr
+  kind: PlacementRule
+  apiGroup: apps.open-cluster-management.io
+subjects:
+- name: test-op-err-initial
+  kind: Policy
+  apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
```
yiraeChristineKim, line 44: This is one deprecated. Next time, using placement is good. For now ok!!

test/integration/operator_policy_error_test.go
JeffeyL: These would have to be added into the actual controller code AFAIK, @JustinKuli any thoughts on this?

JustinKuli: If I had done https://issues.redhat.com/browse/ACM-9896 correctly the first time, this would have been obvious sooner, but `namespace` shouldn't be required in the OperatorPolicy, when it's deployed on OpenShift like it is in these tests. So this test case will need to be updated (possibly just removed).

JustinKuli (dismissed):

test/integration/operator_policy_error_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		AfterAll(func() {
+			_, err := common.OcHub(
+				"delete", "-f", policyPath,
+				"-n", userNamespace, "--ignore-not-found=true",
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = common.OcManaged(
+				"delete", "ns", subscriptionNamespace,
```
dhaiducek, line 37: In the unlikely event the first test fails, this is also going to fail. It's probably safer to add an `--ignore-not-found` here also.

test/resources/operator_policy/test-op-err-correct.yaml
```diff
@@ -0,0 +1,49 @@
...
+metadata:
+  name: test-op-err-initial-plr
+spec:
+  clusterSelector:
+    matchExpressions: []
```
dhaiducek, line 49: Missing trailing newline

test/resources/operator_policy/test-op-err-initial.yaml
```diff
@@ -0,0 +1,49 @@
...
+  kind: Policy
+  apiGroup: policy.open-cluster-management.io
+---
+apiVersion: apps.open-cluster-management.io/v1
+kind: PlacementRule
+metadata:
+  name: test-op-err-initial-plr
+spec:
+  clusterSelector:
+    matchExpressions: []
```
dhaiducek, line 49: Missing trailing newline

dhaiducek (changes requested):
Overall looks great! Some minor cleanup things and then I think it's good to merge.

An aside: This sort of repeated flow with the policy being reset ought to lend itself to parallelism and separating out each one into a unit with a unique policy name that can be checked independently, but that's an exploration/tech debt for another day. 🙂

dhaiducek (approved):

PR #746, opened 2024-03-05
OperatorPolicy test fixes

dhaiducek:
- Check the message before compliance (since that test failure could be more helpful)
- Fetch the CSV name dynamically (since a static version could be flaky if a new z-stream is released)
- Only attempt an OperatorGroup/CSV deletion if a name was discovered
- Clean up all CSV when deployed clusterwide (this is mainly for the benefit of the Canaries)

Followup to:
- #743 
- #736

### DIFF
--- a/test/integration/policy_install_operator_test.go
+++ b/test/integration/policy_install_operator_test.go
@@ -4,8 +4,12 @@
 package integration
 
 import (
+	"strings"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


dhaiducek:
Looks like 2 `Disabled` events were generated:
```
   [FAILED] Timed out after 30.000s.
  The function passed to Eventually failed at /go/src/github.com/stolostron/governance-policy-framework/test/integration/compliance_history_test.go:338 with:
  expected cluster local-cluster to have one disabled events
  Expected
      <[]interface {} | len:2, cap:2>: [ 
```
FYI @mprahl

dhaiducek:
CI is passing here now! 🎉

gparvin (approved):

PR #748, opened 2024-03-11
Adjust offset for `GetWithTimeout`

dhaiducek:


yiraeChristineKim (approved):
Changing to ExpectWithOffset is great!!

PR #749, opened 2024-03-13
Expose kubeconfig files in hosted mode tests for OCM PRs

zyjjay:
Repositories in OCM utilizes kubeconfig files
with the e2e suffix when deploying controllers.


Makefile
```diff
@@ -371,8 +371,10 @@ kind-bootstrap-hosted: kind-install-hosted install-crds install-resources kind-d
 kind-install-hosted: $(ADDON_CONTROLLER)
 	@cd $(ADDON_CONTROLLER) && KIND_VERSION=$(KIND_VERSION) HOSTED_MODE=true ./build/manage-clusters.sh
 	@cp $(ADDON_CONTROLLER)/kubeconfig_cluster1_e2e ./kubeconfig_$(HUB_CLUSTER_NAME)
+	cp ./kubeconfig_$(HUB_CLUSTER_NAME) ./kubeconfig_$(HUB_CLUSTER_NAME)_e2e
```
dhaiducek, line 374: There's an `_e2e` suffix right above this, so I'm not sure this is necessary. Would you point to the error where this is happening?

zyjjay: The error is happening at the "Patch Component Image" job of the workflow here: https://github.com/open-cluster-management-io/governance-policy-framework-addon/actions/runs/8255632781/job/22582549257. 

Specifically, at the `kind-deploy-controller-dev` step where it tries to create resources using the hub/managed kubeconfigs. After looking at it again, I could probably set an environment variable in the framework-addon to point to the kubeconfig file that's already created.

dhaiducek: So the cause is that this PR:
- https://github.com/stolostron/governance-policy-framework-addon/pull/141

Was never submitted to upstream because the framework tests weren't running there, so it wasn't relevant. The `framework-addon` doesn't have this logic, so it's using the wrong code path. Adding the commit to the upstream PR alongside the framework workflow ought to resolve it.

zyjjay: Oh perfect, I'll do that. Thanks!

Makefile
```diff
@@ -5,7 +5,8 @@ LOCAL_BIN ?= $(PWD)/bin
 deployOnHub ?= false
 RELEASE_BRANCH ?= main
 OCM_API_COMMIT ?= $(shell awk '/open-cluster-management.io\/api/ {print $$2}' go.mod)
-DOCKER_URI ?= quay.io/stolostron
+CALLER_REPO ?= stolostron # default repo to pull from
+DOCKER_URI ?= quay.io/$(CALLER_REPO)
```
dhaiducek, line 9: The `DOCKER_URI` has to be slightly different because the upstream is `quay.io/open-cluster-management` while the GitHub repo is `open-cluster-management-io`.

.github/workflows/kind.yml
```diff
@@ -7,6 +7,10 @@ on:
         required: false
         type: string
         default: 'false'
+      caller_repo:
+        required: false
+        type: string
+        default: 'stolostron'
```
dhaiducek, line 13: You should be able to pull this in dynamically rather than having to set it. (Check out the `github.event` object similar to the `RELEASE_BRANCH`  in the `env`.)

dhaiducek (changes requested):

Makefile
```diff
@@ -4,13 +4,20 @@ PWD := $(shell pwd)
...
   VERSION_TAG ?= latest
 endif
+
+OCM_API_COMMIT ?= $(shell awk '/open-cluster-management.io\/api/ {print $$2}' go.mod)
+CALLER_REPO ?= stolostron
+ifeq ($(CALLER_REPO), open-cluster-management-io)
+  DOCKER_URI ?= quay.io/open-cluster-management
+else
+  DOCKER_URI ?= quay.io/stolostron
+endif
```
dhaiducek, line 19: Nit: Might as well not have the else.
```suggestion
DOCKER_URI ?= quay.io/stolostron
ifeq ($(CALLER_REPO), open-cluster-management-io)
  DOCKER_URI ?= quay.io/open-cluster-management
endif
```

.github/workflows/kind.yml
```diff
@@ -102,6 +102,8 @@ jobs:
         deployOnHub: ${{ matrix.deployOnHub }}
         HOSTED: ${{ matrix.hosted }}
       run: |
+        export CALLER_REPO=${{ github.event.organization.login }}
+
```
dhaiducek, line 106: Nit: If the `github` context is available there, I think this would be cleaner in the `env` rather than as a separate export.
```suggestion
        deployOnHub: ${{ matrix.deployOnHub }}
        HOSTED: ${{ matrix.hosted }}
        CALLER_REPO: ${{ github.event.organization.login }}
      run: |
```

dhaiducek (commented):
Looking great! Some optional comments to consider.

.github/workflows/kind.yml
```diff
@@ -101,7 +101,9 @@ jobs:
       env:
         deployOnHub: ${{ matrix.deployOnHub }}
         HOSTED: ${{ matrix.hosted }}
+        CALLER_REPO: ${{ github.event.organization.login }}
       run: |
+
```
dhaiducek, line 106: Nit: extra newline
```suggestion
```

dhaiducek (approved):
Nice! Let's see what happens! 😄 🤞

PR #755, opened 2024-03-18
Fix more broken URLs in Makefile

zyjjay:


zyjjay:
Not sure if this is the best workaround so open to ideas!

dhaiducek:
> Not sure if this is the best workaround so open to ideas! The managementaddon crds seems like it would always be consistent across the two repos in ocm and Stolostron; I'd be ok with just referencing Stolostron to apply the crds unless I missed something somewhere.

Referencing `stolostron` here seems like the way to go for now. I think tt's effectively the same as the content from the cloned repo, anyway.

PR #757, opened 2024-03-18
Conditionally skip iam and cert policy tests

zyjjay:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,8 @@ DEBUG_DIR ?= test-output/debug
 # Test configuration
 TEST_FILE ?=
 TEST_ARGS ?=
+UPSTREAM_TEST ?= false
+
 ifdef TEST_FILE
 	TEST_ARGS += --focus-file=$(TEST_FILE)
 endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -277,6 +279,9 @@ MANAGED_CLUSTER_NAMESPACE ?= $(MANAGED_CLUSTER_NAME)
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
+	if [ "$(UPSTREAM_TEST)" = "true" ]; then\
```
JustinKuli, line 282: I usually reference this styleguide for bash: https://google.github.io/styleguide/shellguide.html#s6.3-tests, which would suggest
```suggestion
	if [[ "$(UPSTREAM_TEST)" == "true" ]]; then\
```
But I think what you have here will work... And it looks like the single bracket syntax is used somewhere else in this file, so it's up to you if you want to change it in all the places or keep it like this.

JustinKuli (approved):
/hold for optional suggestion

Easiest way to test this is probably to merge it and just try :)

Makefile
dhaiducek: @JustinKuli The default shell is `sh` for Make, so double brackets won't work since we're not setting the shell otherwise.

JustinKuli: Oh, wow, good to know

zyjjay:
Going ahead and merging this!

PR #773, opened 2024-04-02
Fix cert and operator policy in compliance history

yiraeChristineKim:
Fix this https://redhat-internal.slack.com/archives/CVAB9H3DE/p1711973234038419

### DIFF
--- a/test/integration/compliance_history_test.go
+++ b/test/integration/compliance_history_test.go
@@ -565,9 +565,10 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API", 
 
 	It("Creates a policy with a compliant and non-compliant cert policy", func(ctx context.Context) {
 		const policyName = "cert-policy"
-		const certNs = "cert-policy-test-ns"
+		const certNs = "ch-cert-policy-test-ns"
 		const certPath = "../resources/compliance_history/cert-policy.yaml"
 		const certSecret = "cert-secret"
+		const nsPolicyPath = "../resources/compliance_history/cert-prereq.yaml"
 
 		now := time.Now().Format(time.RFC3339Nano)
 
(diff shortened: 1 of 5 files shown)


mprahl:
@yiraeChristineKim I think you'll have the same issue I fixed in the Gatekeeper compliance history tests where we need to use policy to do the clean up since the tests apply to all managed clusters which is an unknown amount in a BVT run.

mprahl:
> > @yiraeChristineKim I think you'll have the same issue I fixed in the Gatekeeper compliance history tests where we need to use policy to do the clean up since the tests apply to all managed clusters which is an unknown amount in a BVT run.
> 
> Ok... but how other tests are run?

I think most are scoped to just a self-managed hub (e.g. local-cluster) and optionally one or the managed cluster. The compliance history tests started using `verifyPolicyOnAllClusters` with placements that apply to all clusters so that our coverage of environments is better in BVT and/or QE runs of our tests.

test/integration/compliance_history_test.go
```diff
@@ -690,21 +693,23 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API",
 			Expect(err).ToNot(HaveOccurred())
 
 			// Delete subscription, catalog, etc at once
-			By("Deleting the namespace")
-			_, err = common.OcHub(
+			By("Deleting the namespace policy")
```
mprahl, line 696: I think the prune object behavior of the policy handles this now.

test/integration/compliance_history_test.go
```diff
@@ -690,21 +693,23 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API",
...
 			)
 			Expect(err).ToNot(HaveOccurred())
 		})
 
 		By("Creating the namespace")
-		_, err := common.OcHub(
-			"create",
-			"ns",
-			opNs,
+		_, err := common.OcManaged(
```
mprahl, line 708: Did you mean to apply this to the hub since this is a policy?

yiraeChristineKim: oh mistake Thanks

test/integration/compliance_history_test.go
```diff
@@ -690,21 +693,23 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API",
...
```
mprahl, line 708: @yiraeChristineKim this still needs to be updated to `OcHub`.

yiraeChristineKim: why?!!

mprahl: Why would you apply a policy on the managed cluster?

mprahl (approved):

PR #775, opened 2024-04-03
Increase verify policy on all clusters func timeout

yiraeChristineKim:
Increase verifyPolicyOnAllClusters timeout from 30 secs to 60 secs as we discussed

Ref: https://redhat-internal.slack.com/archives/CUS1VB8P3/p1712154840798339


dhaiducek (dismissed):

mprahl (changes requested):
@yiraeChristineKim did you mean to change the call verifyPolicyOnAllClusters instead? The timeout is configurable in its arguments.

dhaiducek (approved):

mprahl (approved):

PR #783, opened 2024-04-10
Add CRD to operator-policy message

JustinKuli:
This was added to the compliance message in
https://github.com/open-cluster-management-io/config-policy-controller/pull/222

### DIFF
--- a/test/integration/compliance_history_test.go
+++ b/test/integration/compliance_history_test.go
@@ -188,7 +188,7 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API", 
 			g.Expect(resp.StatusCode).To(
 				Equal(http.StatusOK), "expected the compliance API to return the 200 status code",
 			)
-		}, common.DefaultTimeoutSeconds*2, 1).Should(Succeed())
+		}, common.DefaultTimeoutSeconds*4, 1).Should(Succeed())
 	})
 
 	AfterAll(func(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


dhaiducek (dismissed):

mprahl (dismissed):

JustinKuli:
Hmm, looks like a consistent different issue (just before the thing I thought I fixed)... I'll keep looking.

yiraeChristineKim (approved):

test/integration/compliance_history_test.go
```diff
@@ -614,6 +615,9 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test the compliance history API",
 		)
 		Expect(err).ToNot(HaveOccurred())
 
+		By("Verifying that the policy has created the namespace")
+		verifyPolicyOnAllClusters(ctx, policyNS, prereqPolicyName, "Compliant", defaultTimeoutSeconds)
```
dhaiducek, line 619: Well. This seems important. 😆

dhaiducek (dismissed):
Thanks! Feels like those prereq policy checks ought to resolve a fair bit of the flakiness... 🤞

JustinKuli:
/hold

I've got one other minor change to add here, which will double as a consistency check for the tests. So I will push that after this run (hopefully) passes.

test/integration/compliance_history_test.go
mprahl: Do you want to bump the timout to `defaultTimeoutSeconds *2`? Below as well?

JustinKuli: I expect it to be fine at 30 seconds. The config policy controller just needs to get to the policy once. If we have to use `default * 2` for everything, maybe we should just increase the default :laughing:

yiraeChristineKim (approved):

yiraeChristineKim:
Oh good this fix will fix my PR CI

JustinKuli:
It was a different looking error, so I'd prefer to handle it in a separate PR:
```
Summarizing 1 Failure:
  [FAIL] GRC: [P1][Sev1][policy-grc] Test install Operator When no OperatorGroup is specified [It] Should verify CSV details [BVT]
  /go/src/github.com/stolostron/governance-policy-framework/test/integration/policy_install_operator_test.go:220
```

yiraeChristineKim:
Cherry pick to 2.10?

PR #784, opened 2024-04-12
Remove diff from output

dhaiducek:
The diff will still be in the workflow logs and summary, but it's too long for things like Slack messages.

### DIFF
--- a/build/codebase-check.sh
+++ b/build/codebase-check.sh
@@ -19,7 +19,7 @@ ciopDiff() {
 	CI_OP_DIFF=$(diff ${CI_OP_PATH} ${REPO_CI_OP_PATH})
 	if [[ -n "${CI_OP_DIFF}" ]]; then
 		echo "****"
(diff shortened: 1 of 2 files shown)


build/codebase-check.sh
```diff
@@ -170,11 +170,16 @@ rc=0
 ARTIFACT_DIR=${ARTIFACT_DIR:-${PWD}}
 ERROR_FILE_NAME="codebase-errors.log"
 ERROR_FILE="${ARTIFACT_DIR}/${ERROR_FILE_NAME}"
+SUMMARY_FILE="${ARTIFACT_DIR}/summary-${ERROR_FILE_NAME}"
+OUTPUT_FILES="${ERROR_FILE} ${SUMMARY_FILE}"
```
JustinKuli, line 174: How did I not realize you could `tee` to multiple files? This is great!

JustinKuli (approved):
LGTM

build/codebase-check.sh
dhaiducek: I know! I needed it to go to both files and I was like, "You know what would be great..." and `tee` had already solved my problem! 😄

PR #792, opened 2024-04-18
Add `import-managed.sh` to prow test flow

dhaiducek:


build/import-managed.sh
```diff
@@ -0,0 +1,86 @@
...
+kind: KlusterletAddonConfig
+metadata:
+  name: ${MANAGED_CLUSTER_NAME}
+  namespace: ${MANAGED_CLUSTER_NAME}
+spec:
+  applicationManager:
+    enabled: true
+  certPolicyController:
+    enabled: true
+  iamPolicyController:
```
yiraeChristineKim, line 64: Maybe we don't need this `iamPolicyController`

dhaiducek: Good point!

yiraeChristineKim (approved):

build/import-managed.sh
dhaiducek: Actually, looks like it's required in the KlusterletAddonConfig CRD, FYI:
```
The KlusterletAddonConfig "managed-1" is invalid: spec.iamPolicyController: Required value
```

PR #797, opened 2024-04-23
Use the matching release branch on the policy-collection

mprahl:
This will allow the IAM policy tests to still work on 2.10 and earlier.

Relates:
https://issues.redhat.com/browse/ACM-10859
https://github.com/stolostron/canary-scripts/pull/297
https://github.com/stolostron/governance-policy-framework/pull/796

### DIFF
--- a/test/integration/policy_certificate_test.go
+++ b/test/integration/policy_certificate_test.go
@@ -32,10 +32,10 @@ var _ = Describe(
 	func() {
 		const (
 			policyCertificateName   = "policy-certificate"
-			policyCertificateURL    = policyCollectSCURL + policyCertificateName + ".yaml"
 			expiredCertSecretName   = "expired-cert"
 			policyCertificateNSName = "policy-certificate"
 		)
+		policyCertificateURL := policyCollectSCURL + policyCertificateName + ".yaml"
 
 		It("stable/"+policyCertificateName+" should be created on the Hub", func() {
 			By("Creating the policy on the Hub")
--- a/Makefile
+++ b/Makefile
@@ -356,7 +356,7 @@ e2e-debug-dump:
 
 .PHONY: integration-test
 integration-test: e2e-dependencies
-	$(GINKGO) -v $(TEST_ARGS) test/integration -- -cluster_namespace=$(MANAGED_CLUSTER_NAMESPACE) -k8s_client=$(K8SCLIENT) -is_hosted=$(IS_HOSTED) -cluster_namespace_on_hub=$(MANAGED_CLUSTER_NAMESPACE) -patch_decisions=false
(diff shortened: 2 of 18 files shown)


dhaiducek (dismissed):
/hold (in case we want to update the variable name)
This could potentially have a more generic name like "release-branch" in case there were use cases down the road other than policy-collection.
(Alternatively we could hard-code 2.10 and before and instead of having a variable.)

mprahl:
> The more I think about it, the more I'm thinking for simplicity we hardcode the policy-collection branches for 2.10, 2.9, and 2.8, but use this PR for updates from now into the future.

Yeah, I think let's make it dynamic for 2.11+ but static for <= 2.10

mprahl:
> Thanks Matt. I hope this will fix 2.9 bvt test
> 
> ```
>  [FAILED] Unexpected error:
>       <*errors.errorString | 0xc000297620>: 
>       error running command 'kubectl apply -f policy-role.yaml -n policy-test --kubeconfig=../../kubeconfig_hub':
>        error: the path "policy-role.yaml" does not exist
>       : exit status 1
>       {
>           s: "error running command 'kubectl apply -f policy-role.yaml -n policy-test --kubeconfig=../../kubeconfig_hub':\n error: the path \"policy-role.yaml\" does not exist\n: exit status 1",
>       }
>   occurred
> ```
> 
> Test failed

It does seem like `BeforeSuite` is not called when I expected it. I'll fix it.

test/integration/policy_certificate_test.go
```diff
@@ -32,11 +32,16 @@ var _ = Describe(
...
-			policyCertificateURL    = policyCollectSCURL + policyCertificateName + ".yaml"
 			expiredCertSecretName   = "expired-cert"
 			policyCertificateNSName = "policy-certificate"
 		)
 
+		var policyCertificateURL string
+
+		BeforeAll(func() {
+			policyCertificateURL = policyCollectSCURL + policyCertificateName + ".yaml"
+		})
```
dhaiducek, line 43: This feels a little odd. Why not just move the variable declarations to the `init()` function?

mprahl: Are the flags parsed by then?

mprahl: I just double checked and the flags are not parsed by then which is why I opted for BeforeSuite but BeforeSuite doesn't run until after the tests are detected by Ginkgo which leads to the BeforeAll.

mprahl: @dhaiducek I think it used to be an issue but now it seems I can call `flag.Parse` outside of Ginkgo.

dhaiducek: Seems the "proper" way is to put it in `TestIntegration()`? https://onsi.github.io/ginkgo/#overriding-ginkgos-command-line-configuration-in-the-suite

Does that work? It looks like `go test` might also call `flag.Parse()`, so I'm not sure of the effect of it being called twice...

mprahl: Good find. It did seem to be okay with `flag.Parse()` being called twice but I'll try what you suggested.

mprahl:
@dhaiducek and @yiraeChristineKim okay, the CI passed this time!

dhaiducek (approved):

PR #809, opened 2024-05-14
Fix Slack report filter

dhaiducek:
Removes filter for "It" node types, allowing After/Before nodes to also be reported.

The empty text from https://redhat-internal.slack.com/archives/CUS1VB8P3/p1715427328187239 would now be:
```
- [AfterSuite]
  
  Timed out after 180.000s.
    Namespace grc-e2e-policy-generator should be deleted.
    Expected
        <bool>: false
    to be true
```


JustinKuli (approved):
Nit: change PR title to "_Improve_ Slack report filter" :laughing:

PR #811, opened 2024-05-14
Simplify wait commands

dhaiducek:
`jsonpath` wasn't working (I think perhaps because the CLI in the CI is older), but this looks cleaner anyway.


JustinKuli (approved):
I thought that `kubectl wait` could always only be used on conditions, weird.

PR #813, opened 2024-05-20
Integration test enhancements/fixes

dhaiducek:
Among some minor changes, the major updates here are:
- Remove any remaining `ManagedClusterConditionAvailable` (this might cut down on Canary flakes also!)
- Don't patch placement in Prow (this is unnecessary since it's a full ACM install, and it was also getting in the way of running tests with a managed cluster from proceeding properly)

### DIFF
--- a/test/integration/policy_multiline_templatization_test.go
+++ b/test/integration/policy_multiline_templatization_test.go
@@ -93,19 +93,19 @@ var _ = Describe(
 				} {
 					By(fmt.Sprintf("Checking edited ConfigMap %s/%s on the %s cluster", userNamespace, name, cluster))
 					Eventually(
-						func() string {
+						func() (map[string]string, error) {
 							configMap, err := client.CoreV1().ConfigMaps(userNamespace).Get(
 								ctx, name, metav1.GetOptions{},
 							)
-							if err != nil {
-								return ""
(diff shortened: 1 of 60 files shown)


test/integration/policy_multiline_templatization_test.go
```diff
@@ -128,17 +128,19 @@ var _ = Describe(
...
-							return configMap.Data["extraData"]
+							return configMap.Data, err
 						}, defaultTimeoutSeconds*2, 1,
-					).Should(Equal("exists!"))
+					).Should(
+						HaveKeyWithValue("extraData", "exists!"),
+						fmt.Sprintf("ConfigMap %s/%s on the %s cluster should have expected data",
+							configNamespace, name, cluster,
+						),
+					)
```
JustinKuli, line 143: I need to get in the habit of using descriptions like this, they're super helpful!

JustinKuli (approved):
LGTM

test/integration/policy_multiline_templatization_test.go
dhaiducek: Same! I feel like I'm slowly figuring out some better ways to test with Ginkgo, and this is definitely one of them.

PR #815, opened 2024-05-22
Use `main` branch for `clusteradm`

dhaiducek:
This will give us the absolute latest commit for `clusteradm` now that we're deploying it on a daily basis. Otherwise the fix to `clusteradm` we just merged won't actually show as resolved unless another release comes out, which isn't ideal.

### DIFF
--- a/.github/workflows/repo-config.yml
+++ b/.github/workflows/repo-config.yml
@@ -103,10 +103,10 @@ jobs:
     steps:
       - name: Install clusteradm
         run: |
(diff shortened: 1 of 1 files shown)


JustinKuli (approved):
I guess the latest tag is guaranteed to be installable?

dhaiducek:
> I guess the latest tag is guaranteed to be installable?

Not necessarily, but if there was a released tag, I'd hope it'd work! 😬

PR #825, opened 2024-06-13
Create the 2.12 branches

mprahl:
This includes a few fixes.


CURRENT_SUPPORTED_VERSIONS
```diff
@@ -1,3 +1,4 @@
+2.11
 2.10
 2.9
 2.8
```
dhaiducek, line 4: Keeping 2.8 around is right, but doesn't really _feel_ right 😕

dhaiducek (approved):
Thanks for the updates! I've held it until the Prow PRs merge, since we'll want those configurations in place before the branches are created.

CURRENT_SUPPORTED_VERSIONS
mprahl: Yeah, the script removed it I believe but it needs to stay a little while longer.

mprahl:
@dhaiducek I added the Policy Generator to `repos.txt`.

dhaiducek (approved):
/unhold

dhaiducek:
Let's do this! 😄

PR #831, opened 2024-06-25
Skip checking missing `pipeline` branch

dhaiducek:
`release-2.12` doesn't exist yet.


dhaiducek:
I confirmed that this should also resolve the Slack errors since we breached the 3,000 character limit.

ref: https://api.slack.com/reference/block-kit/blocks#section

mprahl (approved):

PR #834, opened 2024-06-26
Add `must-gather` to release flows

dhaiducek:
- Remove references to grc-ui
- Consolidate the release update into a single PR
- Add the `git-fast-forward` step that must-gather uses
- Add `must-gather` to CI checks

ref: https://issues.redhat.com/browse/ACM-12429

### DIFF
--- a/build/branch-create/repo.txt
+++ b/build/branch-create/repo.txt
@@ -6,3 +6,4 @@ stolostron/governance-policy-framework
 stolostron/governance-policy-framework-addon
 stolostron/governance-policy-propagator
 stolostron/policy-generator-plugin
+stolostron/must-gather
--- a/README.md
+++ b/README.md
@@ -22,7 +22,6 @@ View the following functions of the policy framework:
 
 The governance policy framework consists of following components:
 
-- [Governance dashboard](https://github.com/stolostron/grc-ui): Console
 - Govenance policy framework: A framework to distribute various supported policies to managed clusters and collect results to be sent to the hub cluster.
     - [Policy propagator](https://github.com/stolostron/governance-policy-propagator)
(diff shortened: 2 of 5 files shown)


gparvin:
Does this mean grc owns must-gather now? I think it needs an owner.

gparvin (approved):

gparvin:
For further reviews
/hold

dhaiducek:
> Does this mean grc owns must-gather now? I think it needs an owner.

Sort of. With the Konflux migration tasks, I volunteered to handle `must-gather` and proposed our squad own the CI while CEE own the code. I'm adding my GitHub ID to the OWNERS file.

Ref: https://redhat-internal.slack.com/archives/C04L50S5XM4/p1719339184244149

build/branch-create/repo.txt
```diff
@@ -6,3 +6,4 @@ stolostron/governance-policy-framework
 stolostron/governance-policy-framework-addon
 stolostron/governance-policy-propagator
 stolostron/policy-generator-plugin
+stolostron/must-gather
```
gparvin, line 9: I think if we now own must-gather, you would need to add https://github.com/stolostron/backplane-must-gather repo as well.

dhaiducek: I checked with Disaiah on this. We don't actively keep track of `backplane-X.Y` versions, so I asked him to continue to own the build there. I'll create a TODO to have CI mirror code from `must-gather` so that `must-gather` can be the source of truth and `backplane-must-gather` can be strictly for builds. I think that might make the most sense rather than doing something like setting up `backplane-must-gather` as a submodule or something.

gparvin (approved):

dhaiducek:
/unhold

must-gather is in a reasonably stable state, and I don't think additional reviews are necessary here.

PR #854, opened 2024-08-09
Collect memory in the performance script

mprahl:
See each commit for details.

### DIFF
--- a/test/performance/performance.go
+++ b/test/performance/performance.go
@@ -125,7 +125,7 @@ func getMetrics(
 	insecure bool,
 ) metricData {
 	// print metrics
-	ts, avgController := query(
+	ts, controllerCPUAvg := query(
 		thanosHost,
 		token,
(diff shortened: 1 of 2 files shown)


test/performance/performance.go
```diff
@@ -223,18 +162,65 @@ func getMetrics(
 		insecure,
 	)
 
+	_, controllerMemAvg := query(
+		thanosHost,
+		token,
+		fmt.Sprintf(
+			"sum(avg_over_time(container_memory_working_set_bytes{container='config-policy-controller'}[%dm:30s])) "+
```
gparvin, line 169: Optional: Since this test appears to require running on the hub, it might be nice if we could get propagator stats too.  A future goal would be to simulate multiple managed clusters and policies deployed to them.

mprahl:
Merging since it's just a helper script to perform a performance test. It's not an automated test and rarely used.

PR #759, opened 2024-03-20
Fix Gatekeeper operator cleanup

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-10556

There were some competing causes to the flakiness for these tests:
- We ran into an unrecoverable OLM caching bug outlined in https://access.redhat.com/solutions/6991414, solved by restarting the OLM pods
- The compliance history Gatekeeper cleanup Policy was not getting deleted, causing it to linger and continuously clean up when the Gatekeeper test would run (possibly causing the unrecoverable state?)
- Cleanup commands in Gatekeeper were also getting skipped because not all resources had been deployed and it didn't tolerate that.

The cleanup script was also only cleaning up in certain namespaces when the Gatekeeper cleanup policy had been deployed to the global-set namespace. This updates the commands to clean up policies aggressively across all namespace.

PR #855, opened 2024-08-16
Simplify the Gatekeeper uninstall

mprahl:
Only the "real" CSV in the subscription namespace needs to be deleted. The others are copies that get cleaned up automatically. This should make the test faster and more stable.

### DIFF
--- a/test/resources/compliance_history/policy-uninstall-gk.yaml
+++ b/test/resources/compliance_history/policy-uninstall-gk.yaml
@@ -10,10 +10,9 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
         metadata:
-          name: compliance-api-uninstall-gk-sub
+          name: compliance-api-uninstall-gk
         spec:
           object-templates-raw: |
-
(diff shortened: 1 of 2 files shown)


dhaiducek:
FYI This partially reverts:
- https://github.com/stolostron/governance-policy-framework/pull/759

We were hitting an OLM bug. It seems the bug has been closed in the meantime, so I think it's okay to try it out again.

test/resources/compliance_history/policy-uninstall-gk.yaml
```diff
@@ -10,10 +10,9 @@ spec:
         apiVersion: policy.open-cluster-management.io/v1
         kind: ConfigurationPolicy
         metadata:
-          name: compliance-api-uninstall-gk-sub
+          name: compliance-api-uninstall-gk
         spec:
           object-templates-raw: |
-
```
dhaiducek, line 16: (Optional) I personally like this newline because a quirk makes VSCode render the raw template as YAML rather than as a string, but it's fine to remove.

yiraeChristineKim (approved):
/hold for others

dhaiducek (approved):
/unhold

PR #861, opened 2024-08-29
Fix BVT failures due to placements being removed in the policy collec…

yiraeChristineKim:
The placement related manifests in policy-collection were removed since those are environment specific. This has led to the BVT/SVT tests to fail that relied on those. This task is create the now missing placements and bindings in those tests. Ref: https://issues.redhat.com/browse/ACM-13584
Signed-off-by: yiraeChristineKim <yikim@redhat.com>


dhaiducek (commented):
This doesn't feel comfortable to me. We shouldn't be deleting a default installed namespace. It'll create havoc if our tests are run alongside other tests in places like the canaries.

Instead, I'd prefer that the `policy-test` namespace is used as normal and is bound to the `global` ManagedClusterSet.

dhaiducek:
> This doesn't feel comfortable to me. We shouldn't be deleting a default installed namespace. It'll create havoc if our tests are run alongside other tests in places like the canaries.
> 
> Instead, I'd prefer that the `policy-test` namespace is used as normal and is bound to the `global` ManagedClusterSet.

Oh, I see now that it's not deleted. Still, I think it could be cleaner to have our own namespace that we _can_ delete/manage. 😅

yiraeChristineKim:
> > This doesn't feel comfortable to me. We shouldn't be deleting a default installed namespace. It'll create havoc if our tests are run alongside other tests in places like the canaries.
> > Instead, I'd prefer that the `policy-test` namespace is used as normal and is bound to the `global` ManagedClusterSet.
> 
> Oh, I see now that it's not deleted. Still, I think it could be cleaner to have our own namespace that we _can_ delete/manage. 😅

You mean... should I add managedClusterBinding for our test?

dhaiducek:
> You mean... should I add managedClusterBinding for our test?

Yes, that's what I'm thinking. I think it might be cleaner overall, and it'd behave the same way as `open-cluster-management-global-set` does.

test/integration/integration_suite_test.go
```diff
@@ -157,3 +171,31 @@ func canCreateOpenshiftNamespaces() bool {
 
 	return canCreateOpenshiftNamespacesResult
 }
+
+func applyManagedClusterSetBinding() error {
+	yaml := `
+apiVersion: cluster.open-cluster-management.io/v1beta2
```
mprahl, line 177: Why not use the dynamic client for this?

build/clean-up-cluster.sh
```diff
@@ -38,7 +38,8 @@ function hub() {
             oc delete placements.cluster.open-cluster-management.io -n $ns --all --ignore-not-found
         done
     oc delete ns -l e2e=true --ignore-not-found
-    oc delete ns policy-test duplicatetest --ignore-not-found
+     oc delete ns policy-test --ignore-not-found
```
mprahl, line 41:
```suggestion
    oc delete ns policy-test --ignore-not-found
```

test/common/policy_utils.go
```diff
@@ -69,6 +70,77 @@ func PatchPlacementRule(namespace, name string) error {
...
 }
 
+// ApplyPlacement function creates Placement and PlacementBinding so that it will
+// always only match the targetCluster.
+func ApplyPlacement(namespace, policyName string) error {
+	By("Apply Placement and PlacementBinding" + namespace + "/" +
+		"placement-" + policyName + "/" + "placement-binding-" + "policyName" +
+		" with clusterSelector {name: " + ClusterNamespaceOnHub + "}")
+
+	yaml := `
```
mprahl, line 80: Why not use the dynamic client for this rather than writing to a file?

test/common/policy_utils.go
```diff
@@ -69,6 +70,77 @@ func PatchPlacementRule(namespace, name string) error {
...
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: Placement
+metadata:
+  name: placement-$policyName
+spec:
+  predicates:
+    - requiredClusterSelector:
+        labelSelector:
+          matchExpressions:
+            - { key: name, operator: In, values: ["$clusterNamespaceOnHub"] }
```
dhaiducek, line 103: I think this would be more reliable by using native objects and then applying that using the client, either using the struct directly or using `map[string]interface{}`:
```suggestion
	placementBinding := policiesv1.PlacementBinding{
		ObjectMeta: metav1.ObjectMeta{
			Name: "",
		},
		PlacementRef: policiesv1.PlacementSubject{
			APIGroup: GvrPlacement.Group,
			Kind: "Placement",
			Name: "",
		},
		Subjects: []policiesv1.Subject{
			policiesv1.Subject{
				APIGroup: GvrPolicy.Group,
				Kind: "Policy",
				Name: "",
			},
		},
	}

	placement := clusterv1beta1.Placement{
		ObjectMeta: metav1.ObjectMeta{
			Name: "",
		},
		Spec: clusterv1beta1.PlacementSpec{
			Predicates: []clusterv1beta1.ClusterPredicate{
				clusterv1beta1.ClusterPredicate{
					RequiredClusterSelector: clusterv1beta1.ClusterSelector{
						LabelSelector: metav1.LabelSelector{
							MatchExpressions: []metav1.LabelSelectorRequirement{
								metav1.LabelSelectorRequirement{
									Key: "name",
									Operator: "In",
									Values: []string{""},
								}}}}}}}}
```

dhaiducek (commented):
Thanks for working on this! @mprahl beat me to it, but I think using the client rather than temporary files will be cleaner (especially since things can get weird with permissions when writing files when these tests are run other places).

I also have a musing on whether we could just have a placement and placementBinding initiated up-front and just patch the `subjects` in the binding and then delete them at the end, but in the end I think I like having separate placements since having a single placement and binding could get dicey.

test/integration/integration_suite_test.go
yiraeChristineKim: Dale wants to have a namespace we can control(able to delete). but open-cluster-management-global-set we cannot delete. I think to me both ok. policy-test we use as `Useranmespace`. This is used as constant

test/integration/integration_suite_test.go
```diff
@@ -102,21 +104,33 @@ var _ = BeforeSuite(func(ctx SpecContext) {
 	}
 	Expect(namespaces.Get(ctx, userNamespace, metav1.GetOptions{})).NotTo(BeNil())
 
+	By("Create ManagedClusterSetBinding")
+	err := applyManagedClusterSetBinding()
+	Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
yiraeChristineKim: If I add  in resources directory. Should read file and marshal and replace this right?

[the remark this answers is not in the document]
dhaiducek: No, `oc` accepts file paths.

test/common/policy_utils.go
yiraeChristineKim: great! idea!!

test/integration/integration_suite_test.go
dhaiducek: @mprahl Isn't making a comment here about the namespace. He's making a similar point as I did about using the Golang Kubernetes client rather than creating a temporary file.

yiraeChristineKim: @dhaiducek  you're right!!

test/integration/integration_suite_test.go
```diff
@@ -102,21 +103,33 @@ var _ = BeforeSuite(func(ctx SpecContext) {
...
+
 	_, err := common.OcHub(
-		"delete", "namespace", userNamespace,
-		"--ignore-not-found",
+		"delete", "pod", "default",
+		"pod-that-does-not-exist", "--ignore-not-found",
 	)
 	Expect(err).ToNot(HaveOccurred())
 
 	_, err = common.OcHub(
```
mprahl, line 130: This should also be in the `if userNamespace != "open-cluster-management-global-set" {` block.

test/integration/policy_imagemanifest_test.go
```diff
@@ -109,13 +109,13 @@ var _ = Describe("GRC: [P1][Sev1][policy-grc] Test "+
 	})
 
 	It("stable/"+policyIMVName+" should detect imageManifestVulns", func() {
-		By("Checking if the status of the subscription config policy is Compliant")
+		By("Checking if the status of the OperatorPolicy is Compliant")
 		Eventually(
 			func() string {
 				cfgPol := utils.GetWithTimeout(
```
mprahl, line 115:
```suggestion
				opPol := utils.GetWithTimeout(
```

test/integration/integration_suite_test.go
yiraeChristineKim: sorry Could you elaborate this?

mprahl (approved):

test/common/policy_utils.go
```diff
@@ -69,6 +69,99 @@ func PatchPlacementRule(namespace, name string) error {
...
+											ClusterNamespaceOnHub,
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
dhaiducek, line 105: (Optional) This would be more compact
```suggestion
										}}}}}}}}},
```

test/common/policy_utils.go
```diff
@@ -69,6 +69,99 @@ func PatchPlacementRule(namespace, name string) error {
...
+	_, err := ClientHubDynamic.Resource(GvrPlacement).Namespace(namespace).Create(
+		ctx, &placement, metav1.CreateOptions{},
+	)
+	if err != nil {
+		return err
+	}
+
+	placementBinding := unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "policy.open-cluster-management.io/v1",
```
dhaiducek, line 117: This should use the GVR
```suggestion
			"apiVersion": GvrPolicy.Group + "/" + GvrPolicy.Version,
```

test/common/policy_utils.go
```diff
@@ -69,6 +69,99 @@ func PatchPlacementRule(namespace, name string) error {
...
+			"kind":       "PlacementBinding",
+			"metadata": map[string]interface{}{
+				"name": "placement-binding-" + policyName,
+			},
+			"placementRef": map[string]interface{}{
+				"name":     "placement-" + policyName,
+				"kind":     "Placement",
+				"apiGroup": "cluster.open-cluster-management.io",
```
dhaiducek, line 125: This should use the GVR
```suggestion
				"apiGroup": GvrPlacement,
```

test/common/policy_utils.go
```diff
@@ -69,6 +69,99 @@ func PatchPlacementRule(namespace, name string) error {
...
+			},
+			"subjects": []interface{}{
+				map[string]interface{}{
+					"name":     policyName,
+					"kind":     "Policy",
+					"apiGroup": "policy.open-cluster-management.io",
```
dhaiducek, line 131: This should use the GVR
```suggestion
					"apiGroup": GvrPolicy,
```

test/resources/verify_metrics/noncompliant.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: policy.open-cluster-management.io/v1
 kind: Policy
 metadata:
-  name: policy-verify-metrics-noncompliant
+  name: policy-vm-non
```
dhaiducek, line 4: I'm not a huge fan of using nonstandard acronyms like this, but there are a lot of them at this point from the global-set namespace update and they do shorten the names a fair bit, so we can leave them as-is.

dhaiducek (commented):
Looks great! I have some minor comments to use GVR consistently.

mprahl (dismissed):

dhaiducek (dismissed):

dhaiducek:
Looks like the compliance history and Gatekeeper tests both install the Gatekeeper operator. My suspicion is that previously they were in different namespaces and didn't collide (or at least as frequently). Now that the namespaces are the same, they're running into conflicts:

> intersecting operatorgroups provide the same apis, there are CRDs present for the operator, no existing operator Deployments

mprahl:
> Looks like the compliance history and Gatekeeper tests both install the Gatekeeper operator. My suspicion is that previously they were in different namespaces and didn't collide (or at least as frequently). Now that the namespaces are the same, they're running into conflicts:
> 
> > intersecting operatorgroups provide the same apis, there are CRDs present for the operator, no existing operator Deployments

Interesting observation @dhaiducek! I'm going to go ahead and merge this and I'll work on a fix. I think you're right that the uninstall is not working as expected.

I'll let @yiraeChristineKim focus on the discovered policies table.

mprahl:

Will be fixed in another PR

mprahl (approved):

PR #910, opened 2025-01-23
Add an `objectSelector` integration test

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-15811

### DIFF
--- a/test/integration/policy_selector_test.go
+++ b/test/integration/policy_selector_test.go
@@ -1,0 +1,203 @@
+// Copyright (c) 2022 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package integration
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+
+	"github.com/stolostron/governance-policy-framework/test/common"
+)
+
+var _ = Describe(
+	"GRC: [P1][Sev1][policy-grc] Test objectSelector and context variables",
+	Ordered,
+	func() {
+		const (
+			policyName         = "policy-selector"
+			policyYAML         = "../resources/policy_selector/policy-selector.yaml"
+			configMapName      = "selector-config"
+			unrelatedConfigMap = "other-config"
+			configNamespace    = "policy-selector-test"
+			selectorKey        = "policy-selector-e2e"
+		)
+		configMapNames := []string{configMapName + "1", configMapName + "2"}
+
+		generateStatus := func(names []string) string {
+			return fmt.Sprintf("NonCompliant; violation - configmaps [%s] "+
+				"found but not as specified in namespace policy-selector-test",
+				strings.Join(names, ", "))
+		}
+
+		BeforeAll(func(ctx SpecContext) {
+			By("The ConfigMaps should be created on the Managed cluster")
+			By(fmt.Sprintf("Create namespace %s on the Managed cluster", configNamespace))
+			namespace := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: configNamespace},
+			}
+
+			_, err := clientManaged.CoreV1().Namespaces().Create(ctx, namespace, metav1.CreateOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			for i, name := range configMapNames {
+				By(fmt.Sprintf("Creating ConfigMap %s/%s on the Managed cluster", configNamespace, name))
+				configMap := &corev1.ConfigMap{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:   name,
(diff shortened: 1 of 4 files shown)


test/integration/policy_selector_test.go
```diff
@@ -0,0 +1,230 @@
...
+			policyYAML         = "../resources/policy_selector/policy-selector.yaml"
+			configMapName      = "selector-config"
+			unrelatedConfigMap = "other-config"
+			configNamespace    = "policy-selector-test"
+			selectorKey        = "policy-selector-e2e"
+		)
+
+		configMapNames := []string{}
+		// Populate the ConfigMap names used in the tests (selector-config1, selector-config2, ...)
+		for i := 1; i <= 2; i++ {
```
JustinKuli, line 37: Will we ever have more than 2? It's not a big deal, but I feel like it'd be simpler to just pre-populate the list.

test/integration/policy_selector_test.go
```diff
@@ -0,0 +1,230 @@
...
+						Name: unrelatedConfigMap,
+					},
+				}, metav1.CreateOptions{})
+				if !k8serrors.IsAlreadyExists(err) {
+					Expect(err).ToNot(HaveOccurred())
+				}
+			}
+		})
+
+		DescribeTableSubtree("Reporting the correct status", baseSelectorTest, testEntries)
```
JustinKuli, line 151: Is there a reason this isn't just `DescribeTable`?

Also, I think it could be more readable if the `testEntries` were defined directly here, and not converted from a map into a list. Like:
```golang
DescribeTable("Reporting the correct status", baseSelectorTest,
	Entry("With empty objectSelector", defaultTest),
	Entry("With single label exists objectSelector",  selectorTest{
		selector: []v1.LabelSelectorRequirement{
			{Key: selectorKey, Operator: "Exists", Values: []string{}},
		},
		status: "NonCompliant; violation - configmaps [selector-config1, selector-config2] " +
			"found but not as specified in namespace policy-selector-test",
	}),
	...
```

test/integration/policy_selector_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+		Describe("Reporting correct status with object updates", func() {
+			// Run a predefined base set of checks, including applying the policy
+			baseSelectorTest(defaultTest)
```
JustinKuli, line 155: Does this mean that default test is run twice? I think that's ok, I guess the side-effects are useful here?

test/integration/policy_selector_test.go
```diff
@@ -0,0 +1,230 @@
...
+						defaultTimeoutSeconds, 1).Should(Equal(fmt.Sprintf("NonCompliant; violation - configmaps "+
+						"[%s] "+
+						"found but not as specified in namespace policy-selector-test",
+						strings.Join(configmaps, ", "))))
+				})
+			}
+
+			// Test creating and deleting ConfigMaps
+			cmChangeTest("create")
+			cmChangeTest("delete")
```
JustinKuli, line 179: Will we need more tests for other CRUD operations? If not, I feel like the function can just be expanded for these two cases and it would be simpler to read.

JustinKuli (commented):
I think the tests look good, I have some suggestions for re-organizing things to maybe make it more readable. But well, you know, that's just like my opinion, man.

test/integration/policy_selector_test.go
dhaiducek: Probably not--it's a fair point. I'll pre-populate instead. 😄

test/integration/policy_selector_test.go
dhaiducek: > Is there a reason this isn't just `DescribeTable`?

The `DescribeTableSubtree` allows nesting `It` nodes (basically wraps in a `Describe` rather than an `It` as `DescribeTable` does: https://onsi.github.io/ginkgo/#generating-subtree-tables

> Also, I think it could be more readable if the `testEntries` were defined directly here, and not converted from a map into a list.

It's true. Originally I was going to use the test name to generate a unique policy name to allow them to run in parallel, but since I didn't do that I agree it makes more sense to just dump them in as arguments.

test/integration/policy_selector_test.go
dhaiducek: Yes--it includes applying the policy. Basically kind of a reset before moving on to other checks.

dhaiducek: Though that said, I could do a refactor to have a `BeforeEach` to apply the policy instead...

test/integration/policy_selector_test.go
dhaiducek: Yeah, there probably won't be other CRUD operations (since updates are pretty well tested elsewhere).

JustinKuli (approved):
LGTM! Thanks for the updates.

I think the failures have just been the CI having trouble getting a cluster, right? If that continues to be an issue, I'd be ok overriding the check if we have evidence the tests passed in a real cluster.

dhaiducek:
> I think the failures have just been the CI having trouble getting a cluster, right? If that continues to be an issue, I'd be ok overriding the check if we have evidence the tests passed in a real cluster.

It's not the cluster--Prow has some syntax highlighting that highlights the pending states while the cluster is installing. It's this:
```
   [FAIL] GRC: [P1][Sev1][policy-grc] Test install Operator When all defaults are used on the subscription [It] test-op-47229 should be created and become compliant [BVT]
  /go/src/github.com/stolostron/governance-policy-framework/test/integration/policy_install_operator_test.go:517 
```

But it's also been a problem for a while. I'm looking into it.

dhaiducek:
/retest

I couldn't get it to fail locally, so I'm not sure what's going on with that failure. I'll get to to re-run at least until it shows proof that the selector test passes...

PR #916, opened 2025-03-25
[release-2.12] Address `oauth2` vuln

dhaiducek:
Address CVE-2025-22868

PR #922, opened 2025-03-27
[release-2.12] Address `crypto` vuln

dhaiducek:
Address CVE-2025-22869


JustinKuli:
Seems like the linter needs to be disabled? Or maybe it somehow lost its config - seems like it has complaints about test files that I would expect it should be ignoring...

dhaiducek:
> Seems like the linter needs to be disabled? Or maybe it somehow lost its config - seems like it has complaints about test files that I would expect it should be ignoring...

Ah, sorry. This one is blocked by this one, where the linter is disabled:
- https://github.com/stolostron/governance-policy-framework/pull/916

I suppose I could combine them... 🤔

JustinKuli (dismissed):

JustinKuli (approved):

PR #926, opened 2025-06-23
Setup the 2.15 branches

JustinKuli:


JustinKuli:
(pushed a no-op to try the prow jobs again)

dhaiducek:
@JustinKuli A reminder the framework tests run on-demand and the image isn't built by default (not sure why it's showing up here, actually). If you want to run the framework tests, you'll need to issue a `/test remaining-required` or else override it.

dhaiducek (approved):

PR #937, opened 2025-07-23
Update arch diagram and desc

jan-law:
ref: https://issues.redhat.com/browse/ACM-10779

Updating the diagram and description to better reflect the current architecture. (The PNG is equivalent to the PNG and SVG files in the internal google drive folder from July 9.)


images/policy-framework-architecture-diagram.png
JustinKuli, line 1: The GitHub interface for comparing images is really cool :sunglasses:

PR #942, opened 2025-09-17
Common Makefile version bumps

dhaiducek:
With the linter updates, also updates the linter configuration file.

I added `//nolint:contextcheck` in some places rather than disabling the linter since I think it's a good linter to have and a good visual reminder kind of like a "TODO", but implementing it (even with AI) would have required more time than reasonable for this initial PR.


JustinKuli (approved):
Looks great!

PR #1047, opened 2026-04-29
Add 5.1 to Current version

yiraeChristineKim:
## Summary by CodeRabbit

* **Chores**
  * Added version 5.1 to the supported versions list alongside the existing version 5.0.


build/branch-create/migrate-konflux.sh
```diff
@@ -9,7 +9,7 @@ set -eo pipefail
 SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
 
 # Read current version from CURRENT_VERSION file
-CURRENT_VERSION=$(cat "${SCRIPT_PATH}/../../CURRENT_VERSION")
+CURRENT_VERSION=$(head -1 "${SCRIPT_PATH}/../../CURRENT_VERSION")
```
dhaiducek, line 12: This is one that I think will need to know about both versions since, if I recall, there's logic here that takes the current version and deletes all other pipeline files.

build/main-branch-sync/sync.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 path="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 exit_code=0
-CURRENT_VERSION=$(cat ${path}/../../CURRENT_VERSION)
+CURRENT_VERSION=$(head -1 ${path}/../../CURRENT_VERSION)
```
dhaiducek, line 6: This one also I'm wondering if we want to go ahead and fast-forward to both branches just to keep them aligned.

PR #1074, opened 2026-07-09
fix: continue on empty commits

dhaiducek:
Empty commits in the bulk update script is causing the script to exit early.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bulk repository updates to handle commit failures gracefully.
  * If a commit does not succeed, the process now skips push and pull request creation for that repository and continues with the next one.
  * Added a clear error message when a commit fails.


jan-law (approved):
LGTM

